PR #6, opened 2017-12-11
Remove Housekeeping implementation from CT

courtsVII:
Phase one in making general purpose Housekeeping project


circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/CircusTrainHousekeepingConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+
+import com.hotels.bdp.circustrain.api.event.HousekeepingListener;
+import com.hotels.housekeeping.repository.LegacyReplicaPathRepository;
+import com.hotels.housekeeping.service.HousekeepingService;
+import com.hotels.housekeeping.service.impl.FileSystemHousekeepingService;
+
+@Configuration
+@ComponentScan("com.hotels.housekeeping")
```
ddcprg: are we all fine with this package name or should this be `com.hotels.bdp.housekeeping` instead?

massdosage: I think it's generally better to avoid putting our team names into package names when going external but with CT itself that would have been a massive backwards incompatible change so we didn't do that there. Here there are probably very few, if any, people using these classes outside of CT so I don't mind changing the package like this.

PR #8, opened 2017-12-13
BDP-1091: Decouple housekeeping

courtsVII:


circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/audit/CircusTrainAuditRevision.java
```diff
@@ -0,0 +1,96 @@
...
+import org.hibernate.envers.RevisionEntity;
+import org.hibernate.envers.RevisionNumber;
+import org.hibernate.envers.RevisionTimestamp;
+
+import com.hotels.housekeeping.audit.AuditRevision;
+
+@Entity
+@RevisionEntity
+@Table(schema = "circus_train", name = "audit_revision")
+public class CircusTrainAuditRevision implements AuditRevision {
```
massdosage: Mmm, the problem with this approach is now that every downstream project that wants to store stuff in a DB will have to replicate all of this code just in order to change the db/table name. Could we not keep this class in the housekeeping module and if someone wants to change the DB name they just extend the class with an empty implementation that overrides the Table annotations? Is there really no other way to make the schema and table names configurable?

circus-train-housekeeping/src/test/java/com/hotels/bdp/circustrain/housekeeping/LegacyReplicaPathRepositoryTest.java
```diff
@@ -0,0 +1,118 @@
+package com.hotels.bdp.circustrain.housekeeping; /**
```
massdosage: Similar here, I think we should have a configurable base DB implementation that can be used more or less "out of the box" in the housekeeping module that has all these tests etc. Otherwise this has to be repeated in each downstream project.

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/audit/CircusTrainAuditRevision.java
courtsVII: The MappedSuperclass approach broke our Auditing, I will need to take some more time to look at it and how Entity Inheritance interacts with Auditing

 I'm sure it is doable with a rethink of how we are wiring things up, I need to spend more time on it, the inheritance stuff also has downsides so despite repeated code this is simple and easy to manage with less side effects and weird Hibernate gotchas, bit of a tradeoff

circus-train-housekeeping/src/test/java/com/hotels/bdp/circustrain/housekeeping/LegacyReplicaPathRepositoryTest.java
ddcprg: why is `package` before the file header?

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/listener/HousekeepingListener.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.hotels.bdp.circustrain.api.event;
```
ddcprg: now that CT is open source we should not make this sort of changes without making deprecating the old class first - this is the API JAR. I would keep the old class, mark it with `@Deprecated` and add comment that this class must be replaced with the copy in the new package. We can then remove the class in a major release

massdosage: Normally I'd agree but we're already doing a major version bump from CT 10.0.0 to CT 11.0.0 when you compare master to the last release. Also, I'm not sure anyone outside of us is actually using CT yet so I wouldn't be as cautious as normal. The change does need to be clearly documented in the changelog though.

circus-train-housekeeping/pom.xml
```diff
@@ -43,10 +53,41 @@
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
+      <version>1.9.5</version>
```
massdosage: version before scope please

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/audit/CircusTrainAuditRevision.java
```diff
@@ -0,0 +1,29 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+import org.hibernate.envers.RevisionEntity;
+
+import com.hotels.housekeeping.audit.EntityAuditRevision;
+
+@Entity
+@RevisionEntity
+@Table(schema = "circus_train", name = "audit_revision")
```
massdosage: OK, so this is how we're allowing people to define the schema name. What happens if we made the parent class non-abstract with a @Table definition on it with a schema called say "housekeeping"? Would we still be able to do what we're doing here to change the name of the table? That way we can really provide a default implementation of housekeeping that "just works" out of the box but still allow customisations like this.

circus-train-housekeeping/src/test/java/com/hotels/bdp/circustrain/housekeeping/JdbcHousekeepingListenerTest.java
```diff
@@ -54,5 +55,4 @@ public void typical() {
     assertThat(argumentCaptor.getValue().getPathEventId(), is(PATH_EVENT_ID));
     assertThat(argumentCaptor.getValue().getPath(), is("/foo/bar"));
   }
-
-}
+}
```
massdosage: New line please

circus-train-housekeeping/src/test/java/com/hotels/bdp/circustrain/housekeeping/LegacyReplicaPathRepositoryTest.java
massdosage: +1

massdosage: Since we have a default "entity" implementation in the housekeeping project I think we should move all the DB CRUD type tests back there.

circus-train-housekeeping/src/test/java/com/hotels/bdp/circustrain/housekeeping/LegacyReplicaPathRepositoryTest.java
```diff
@@ -0,0 +1,118 @@
...
+    "spring.jpa.hibernate.ddl-auto=update",
+    "spring.jpa.hibernate.generate-ddl=true",
+    "spring.datasource.initialize=true",
+    "spring.datasource.schema=classpath:/schema.sql",
+    "spring.datasource.max-wait=10000",
+    "spring.datasource.max-active=5",
+    "spring.datasource.test-on-borrow=true",
+    "housekeeping.expired-path-duration=P3D",
+    "housekeeping.data-source.driver-class-name=org.h2.Driver",
+    "housekeeping.data-source.url=jdbc:h2:./target/data/housekeeping;AUTO_SERVER=TRUE;DB_CLOSE_ON_EXIT=FALSE",
```
massdosage: Is there no way this can be specified programatically so we could use a temp folder and get auto cleanup on it?

circus-train-tool-parent/circus-train-comparison-tool/src/test/resources/log4j.xml
```diff
@@ -44,4 +44,4 @@
     <appender-ref ref="CONSOLE" />
   </root>
 
-</log4j:configuration>
+</log4j:configuration>
```
massdosage: new line

circus-train-tool-parent/circus-train-filter-tool/src/test/resources/test-log4j.xml
```diff
@@ -32,4 +32,4 @@
     <appender-ref ref="CONSOLE" />
   </root>
 
-</log4j:configuration>
+</log4j:configuration>
```
massdosage: ditto

circus-train-tool-parent/circus-train-tool/src/main/resources/log4j.xml
```diff
@@ -39,4 +39,4 @@
     <appender-ref ref="CONSOLE" />
   </root>
 
-</log4j:configuration>
+</log4j:configuration>
```
massdosage: ditto

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/audit/CircusTrainAuditRevision.java
courtsVII: Unfortunately you cannot do that

From the docs:

Mapped Superclasses:
Entities may inherit from superclasses that contain persistent state and mapping information but are not entities. That is, the superclass is not decorated with the @Entity annotation and is not mapped as an entity by the Java Persistence provider. These superclasses are most often used when you have state and mapping information common to multiple entity classes.

Mapped superclasses cannot be queried and can’t be used in EntityManager or Query operations. You must use entity subclasses of the mapped superclass in EntityManager or Query operations. Mapped superclasses can’t be targets of entity relationships. Mapped superclasses can be abstract or concrete.

Mapped superclasses do not have any corresponding tables in the underlying datastore. Entities that inherit from the mapped superclass define the table mappings.

circus-train-housekeeping/src/test/java/com/hotels/bdp/circustrain/housekeeping/LegacyReplicaPathRepositoryTest.java
courtsVII: There is cleanup on it if you look at the teardown() method in this class

massdosage: Yeah, but we've had to write that code and also hardcode a path under the target folder, if we were able to use the location from a JUnit temp folder we'd get all that "for free".

circus-train-housekeeping/pom.xml
```diff
@@ -12,6 +12,11 @@
   <name>Housekeeping</name>
   <description>Maintenance process</description>
 
+  <properties>
```
massdosage: Can these be removed?

circus-train-housekeeping/pom.xml
```diff
@@ -20,20 +25,25 @@
...
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
       <scope>provided</scope>
     </dependency>
 
     <dependency>
-      <groupId>com.hotels</groupId>
-      <artifactId>housekeeping</artifactId>
-      <version>0.0.1</version>
+      <groupId>com.google.guava</groupId>
```
massdosage: Double check whether we need this

circus-train-housekeeping/src/test/java/com/hotels/bdp/circustrain/housekeeping/JdbcHousekeepingListenerTest.java
```diff
@@ -55,4 +56,4 @@ public void typical() {
     assertThat(argumentCaptor.getValue().getPath(), is("/foo/bar"));
   }
 
-}
+}
```
massdosage: newline

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/model/CircusTrainLegacyReplicaPath.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.persistence.UniqueConstraint;
+
+import org.hibernate.envers.Audited;
+
+import com.hotels.housekeeping.model.EntityLegacyReplicaPath;
+
+@Entity
+// @Converter doesn't work with @Audited until https://hibernate.atlassian.net/browse/HHH-9042 is released
+@Audited
+@Table(schema = "circus_train", name = "legacy_replica_path",
```
ddcprg: Ok, ignore my comment on the PR. This is fine but we still need to conditionally load the entities in the other project... or define a default HK profile

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/audit/CircusTrainAuditRevision.java
ddcprg: a condition is also required for this entity in the HK module

PR #11, opened 2017-12-14
Doclint and travis profiles now in parent

massdosage:
I'll replace SNAPSHOT version of parent when HotelsDotCom/hotels-oss-parent#10 is merged and released


ddcprg:
Add a line in CHANGELOG.md

massdosage:
Since the change isn't user facing or affect circus-train itself in any way I didn't think it was worth adding to the changelog.

PR #21, opened 2018-01-26
Addition of location, metastore uris and partition keys to SNS messages

### LINKED ISSUE #20: Improve SNS mesage format
Currently the SNS output from Circus Train is pretty "bare bones" as this was built as a reference implementation. Some suggested changes:
* Make protocolVersion a Semantic Version number.
* Add table base location.
* Add replica Hive metastore URL so downstream process can connect to this to find extra metadata.
* Add partition key name and type for partitioned table replications. These fields must be in same order as the existing partition values are output.

massdosage:
Fixes #20 

Changing version number to 12.0.0 might be more appropriate since I changed some public interfaces. Let's see what survives the review process and I'll change accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 ### Added
 * Added `replicaTableLocation`, `replicaMetastoreUris` and  `partitionKeys` to SNS message.
 ### Changed
-* SNS Message `protocolVersion` changed from "1.0" to "2.0.0".
+* SNS Message `protocolVersion` changed from "1.0" to "1.1".
 * Updated documentation for circus-train-aws-sns module (full reference of SNS message format, more examples).
 * Fixed references to README.md in command line runner help messages to point to correct GitHub locations.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 # Unreleased
+### Changed
+* Fixed references to README.md in command line runner help messages to point to correct GitHub locations.
 
 # 11.0.0 - 2018-01-16
 ### Changed
 * Upgraded Hive version from 1.2.1 to 2.3.2 (changes are backwards compatible).
 * Upgraded Spring Platform version from 2.0.3.RELEASE to 2.0.8.RELEASE.
-* Replaced TunnellingMetaStoreClient "concrete" implementation with a Java reflection TunnellingMetaStoreClientInvocationHandler. 
+* Replaced `TunnellingMetaStoreClient` "concrete" implementation with a Java reflection `TunnellingMetaStoreClientInvocationHandler`. 
 * Replicating a partitioned table containing no partitions will now succeed instead of silently not replicating the table metadata.
 * Most functionality from Housekeeping module moved to https://github.com/HotelsDotCom/housekeeping.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
+### Added
+* Added `replicaTableLocation`, `replicaMetastoreUris` and  `partitionKeys` to SNS message.
 ### Changed
+* SNS Message `protocolVersion` changed from "1.0" to "2.0.0".
+* Updated documentation for circus-train-aws-sns module (full reference of SNS message format, more examples).
 * Fixed references to README.md in command line runner help messages to point to correct GitHub locations.
 
 # 11.0.0 - 2018-01-16
--- a/circus-train-aws-sns/README.md
+++ b/circus-train-aws-sns/README.md
@@ -1,11 +1,12 @@
 # SNS event listener
 
 ##  Overview
-The SNS event listener is an optional Circus Train component which, if found on the classpath while running 
-Circus Train, will push replication events as JSON onto Amazon Web Services SNS topics.
+The SNS event listener is an optional Circus Train component which, if found on the classpath, will push 
+replication events as JSON onto Amazon Web Services SNS topics.
 
 ## Installation
 The listener can be activated in a number of ways which are described below.
+
 ### Classpath
 The jar file produced by this project can be retrieved from Maven Central and stored in a location available 
 when running Circus Train. The code snippet below shows how you could add this to the Circus Train classpath 
@@ -14,7 +15,8 @@ in a script before calling out to the main Circus Train script:
     export CIRCUS_TRAIN_CLASSPATH=$CIRCUS_TRAIN_CLASSPATH:/your-circus-train-sns-lib-path/*
 
 ### Dependency
-If you have a project that is using Circus Train as a Maven dependency you could add the XML fragment below to your POM file in order to depend on the Circus Train SNS artifact and bundle this onto your application's 
+If you have a project that is using Circus Train as a Maven dependency you could add the XML fragment below 
+to your POM file in order to depend on the Circus Train SNS artifact and bundle this onto your application's 
 classpath at runtime:
 
     <dependency>
--- a/circus-train-aws-sns/README.md
+++ b/circus-train-aws-sns/README.md
@@ -1,11 +1,22 @@
 # SNS event listener
 
 ##  Overview
-Push replication events as JSON onto Amazon Web Services SNS topics.
+The SNS event listener is an optional Circus Train component which, if found on the classpath while running 
+Circus Train, will push replication events as JSON onto Amazon Web Services SNS topics.
 
 ## Installation
-Add the connector as a dependency to your Circus Train project or put in on the Circus Train classpath:
+The listener can be activated in a number of ways which are described below.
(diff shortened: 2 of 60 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
+### Added
+* Added `replicaTableLocation`, `replicaMetastoreUris` and  `partitionKeys` to SNS message.
 ### Changed
+* SNS Message `protocolVersion` changed from "1.0" to "2.0.0".
```
ddcprg: I don't think we need an extra number for the versions here, M.m should be sufficient

massdosage: But then we lose the benefits of semantic versioning and that you can just point people to the website and say "go there, this explains how our versions work". I'm all for "sem ver everywhere".

circus-train-aws-sns/README.md
```diff
@@ -53,28 +70,110 @@ Currently the S3 credentials specified in the JCEKS key store are used to authen
...
+	  "replicaCatalog" : "replicaCatalogName",
+	  "sourceTable" : "srcDb.srcTable",
+	  "replicaTable" : "replicaDb.replicaTable",
+	  "replicaTableLocation" : "s3://bucket/path",
+	  "replicaMetastoreUris" : "thrift://host:9083",
+	  "bytesReplicated" : 84837488
+	}
+
+#### Replication success for a partitioned table
+The following shows an example JSON message representing a successful replication of a non-partitioned table
```
ddcprg: -> "successful replication of a partitioned table" ?

massdosage: Copy-pasta error, thanks for spotting.

circus-train-aws-sns/src/test/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListenerTest.java
```diff
@@ -53,15 +56,19 @@
...
   private static final RuntimeException ERROR = new RuntimeException("error message");
   private static final String ENDTIME = "endtime";
-  private static List<EventPartition> PARTITIONS_0;
-  private static List<EventPartition> PARTITIONS_1;
+  private static EventPartition PARTITION_0;
+  private static EventPartition PARTITION_1;
+
   private static final String STARTTIME = "starttime";
   private static final String EVENT_ID = "EVENT_ID";
+  private static final String PROTOCOL_VERSION = "\"protocolVersion\":\"2.0.0\"";
```
ddcprg: can you change this to be just the number and put `"protocolVersion"` back in the string? It harder to read like this

massdosage: Sure, my thinking was this was less repeated text to type but I agree it's not as easy to read and the part that is going to change is the number, not the rest of it. Will change like you suggest.

circus-train-aws-sns/README.md
```diff
@@ -53,28 +70,110 @@ Currently the S3 credentials specified in the JCEKS key store are used to authen
...
+|`replicaCatalog`|String|Only when configured|The configured name of the replicat catalog|
+|`sourceTable`|String|Always|The fully qualified name of the source table|
+|`replicaTable`|String|Always|The fully qualified name of the replica table|
+|`replicaTableLocation`|String|Always|The replicate table location|
+|`replicaMetastoreUris`|String|Always|The Hive metastore URIs of the replica Hive Thrift metastore|
+|`partitionKeys`|Ordered Map of String->String|Only on successful replication of a partitioned table|A map where the keys are the partition key names and the values are the partition key types, the order output in the JSON matches the order of the values in the `modifiedPartitions` field below|
+|`modifiedPartitions`|List of List of Strings|Only on successful replication of a partitioned table|A list containing Lists of string representing the partition key values|
+|`bytesReplicated`|Long|Only on replication of `type` SUCCESS of a table containing data|The number of bytes of data replicated|
+|`errorMessage`|String|Only on replication of `type` FAILURE|A message describing the cause of the failure|
```
patduin: what actually happens if the message is too big for SNS to process, does it get truncated or fails to sent?

CHANGELOG.md
ddcprg: This is a "protocol", there is no patch/build version in protocols

circus-train-aws-sns/README.md
patduin: Ok just read the Limitations bit that explains what happens.

circus-train-aws-sns/README.md
```diff
@@ -53,28 +70,110 @@ Currently the S3 credentials specified in the JCEKS key store are used to authen
...
+	  "replicaTable" : "replicaDb.replicaTable",
+	  "replicaTableLocation" : "s3://bucket/path",
+	  "replicaMetastoreUris" : "thrift://host:9083",
+	  "errorMessage" : "Connection timed out"
 	}
 
 ## Limitations
-SNS messages have a size limit of 256KB. It is possible that you could exceed this bound if your replication touches a very large number of partitions. In this case your message will not be sent and instead logged with a warning.
\ No newline at end of file
+SNS messages have a size limit of 256KB. It is possible that you could exceed this bound if your replication touches a very large number of partitions. In this case your message will not be sent and instead logged with a warning.
```
patduin: So you could have a SUCCESS with a large message that cannot be sent and will be logged, would that then fail the replication?
Trying to figure how a user is supposed to find out if something like this happens. But looks like reading the logs is the only way.

circus-train-aws-sns/README.md
massdosage: That's a good question. I didn't change the sending code, if you look at the publish() method in SNSListener you'll see that we log a warning and then we try send it any way but capture the exception. My guess is that the AWS client/server will throw an exception. I'll take a look a bit later, would be nice to know.

circus-train-aws-sns/README.md
massdosage: I don't think we'll fail the replication so yes, they'd have to look in the logs.

circus-train-aws-sns/src/test/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListenerTest.java
```diff
@@ -115,20 +132,30 @@ public void start() {
...
-            + "\"replicaDb.replicaTable\"}"));
+    assertThat(request.getMessage(), is("{"
+        + PROTOCOL_VERSION
+        + ",\"type\":\"START\",\"headers\":{\"pipeline-id\":\"0943879438\"},"
+        + "\"startTime\":\"starttime\",\"eventId\":\"EVENT_ID\",\"sourceCatalog\":\"sourceCatalogName\","
+        + "\"replicaCatalog\":\"replicaCatalogName\",\"sourceTable\":\"srcDb.srcTable\",\"replicaTable\":"
+        + "\"replicaDb.replicaTable\",\"replicaTableLocation\":\"s3://bucket/path\",\"replicaMetastoreUris\":\"thrift://host:9083\"}"));
   }
 
+  // TODO: need tests for unpartitioned tables to see what gets output for partition stuff
```
patduin: still going todo?

massdosage: No, it's done in successUnpartitionedTable(), will remove the todo, thanks for spotting.

CHANGELOG.md
massdosage: You could make a backwards compatible change to the message that fixes a bug which would result in updating the "hotfix" number. This would be different to making a backwards compatible change to the message that adds a new feature which would result in updating the "minor version" number. So I think you can argue that SemVer is just as useful for API versioning as it is for software versioning.

ddcprg: I disagree, any changes in the format make a new version of the protocol no matter if they are backwards compatible or not

massdosage: So you're disagreeing with the definition of SemVer then and suggesting we don't use it here? I'd prefer to use it as its a well agreed upon, documented and known versioning scheme that we can point to instead of having to explain our own here. Any changes do require a new version, it just determines which of the 3 numbers you change and by having 3 instead of 2 you can communicate the difference between a change due to a bug fix and a change due to the addition of a new feature.

ddcprg: I disagree with having 3 number for protocol versioning

massdosage: OK, anyone else have strong opinions on this? If necessary I'll set up a quick meeting to get a group decision.

patduin: I quite like have the same SemVer versioning scheme in all our version numbers it makes it easier to talk about. I don't really see any harm in using it here.

PR #29, opened 2018-02-08
[issue-28] Upgrade housekeeping

### LINKED ISSUE #28: Upgrade housekeeping
Upgrade to version 1.0.2 of housekeeping

ddcprg:
Fixes #28


circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
```diff
@@ -90,6 +90,7 @@ public static void main(String[] args) throws Exception {
           .properties("instance.home:${user.home}")
           .properties("instance.name:${source-catalog.name}_${replica-catalog.name}")
           .properties("jasypt.encryptor.password:${password:null}")
+          .properties("housekeeping.schema-name:circus_train")
```
ddcprg: This should be done in the housekeeping module but I can't find a cleaner way of setting this property other than this.

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/repository/CircusTrainLegacyReplicaPathRepository.java
```diff
@@ -18,11 +18,10 @@
 import org.springframework.context.annotation.Lazy;
 import org.springframework.transaction.annotation.Transactional;
 
-import com.hotels.bdp.circustrain.housekeeping.model.CircusTrainLegacyReplicaPath;
+import com.hotels.housekeeping.model.HousekeepingLegacyReplicaPath;
```
ddcprg: I think we could get rid of this interface and move the annotations to the base interface, let me know what you think

circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
massdosage: Understood

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/repository/CircusTrainLegacyReplicaPathRepository.java
massdosage: Sounds like a good idea to me, simpler for end users and the base interface is already couled to spring data package via imports.

PR #30, opened 2018-02-09
Fix failing Travis build

### LINKED ISSUE #27: Fix failing Travis build
Circus Train builds in Travis are currently failing with an error message like so:

"No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

for example https://travis-ci.org/HotelsDotCom/circus-train/builds/334657612

We need to investigate why this is happening. Note: we haven't ever had a successful build so perhaps we should first run a more limited version of the build and get this to succeed and then add more steps to try pinpoint where the failure is coming from.

massdosage:
Fixes #27.

This is a bit of a weird one TBH. Summary of changes:
* Remove "travis" profile in pom and tweaked some log4j settings and travis.yml to reduce size of log to less than 4MB but have it visible in the console to ease debugging.
* I found that the DynamicInputFormatTest and SimpleCopyListingTest which we inherited from distcp use a mini hdfs cluster and when shutdown() is called on this it appears to lock the system in Travis. So for the latter I removed the use of the cluster and replaced it directly with a FileSystem. For the former test I just removed the cluster.shutdown() call in the "@After" method which felt a built dirty but it works. I tried shutting it down in an async thread which I ignored the result of this but this still locked up the whole build.


.travis.yml
```diff
@@ -7,9 +7,8 @@ cache:
...
     - ~/.m2/repository
 install:
-  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V -P travis
+  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true --quiet --batch-mode --show-version --activate-profiles travis
 before_script:
   - echo "MAVEN_OPTS=${TEST_MEM_SETTINGS}" > ~/.mavenrc
 script:
-  - mvn clean test cobertura:cobertura coveralls:report -B -P travis -DrepoToken=${COVERALLS_REPO_TOKEN} -Dtest.arguments="${TEST_MEM_SETTINGS}"
-
+  - mvn cobertura:cobertura coveralls:report --batch-mode --activate-profiles travis -DrepoToken=${COVERALLS_REPO_TOKEN} -Dtest.arguments="${TEST_MEM_SETTINGS}"
```
patduin: still need to set the active profile here since you removed it from the pom? same in the 'install' line above

ddcprg: yes, it's still need it... the profile in this pom was only for debugging

massdosage: The "travis" profile is defined in the parent pom so we still need that in order to kick in Coveralls, the bit I removed from this project's pom was an extension of that profile to not log to the console which we no longer need.

massdosage:
@ddcprg there are no customer or end user facing changes in this so I don't think we need to mention it in the changelog. It's all tests and internal build stuff.

PR #50, opened 2018-03-21
Issue 49: S3-S3 Copier Hive Diff Fix

courtsVII:
S3 - S3 copying with generate-partition-filter set to true copies all of the tables partitions every time when it should't.

This occurred cause we were creating checksums which depended on the last modified date of a directory, however in S3 the Hadoop api will return the current time rather than a meaningful value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # TBD
 ### Changed
 * Clearer replica-check exception message. See [#47](https://github.com/HotelsDotCom/circus-train/issues/47).
+### Fixed
+* S3-S3 Hive Diff calculating incorrect checksum on folders
 
 # 11.3.0 - 2018-02-27
 ### Changed
--- a/circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/functions/PathToPathMetadata.java
+++ b/circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/functions/PathToPathMetadata.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Expedia Inc.
+ * Copyright (C) 2016-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # TBD
 ### Changed
 * Clearer replica-check exception message. See [#47](https://github.com/HotelsDotCom/circus-train/issues/47).
+### Fixed
+* S3-S3 Hive Diff calculating incorrect checksum on folders
```
massdosage: Add a See like above with a link to the ticket it fixes (and a period at the end ;))

circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/functions/PathToPathMetadata.java
```diff
@@ -26,6 +26,8 @@
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
```
massdosage: Optimise your imports, I'm assuming these are left over from debugging log statements which have now been removed.

circus-train-comparator/src/test/java/com/hotels/bdp/circustrain/comparator/hive/functions/PathToPathMetadataTest.java
```diff
@@ -38,15 +38,14 @@
...
-import com.hotels.bdp.circustrain.comparator.hive.functions.PathToPathMetadata;
 import com.hotels.bdp.circustrain.comparator.hive.wrappers.PathMetadata;
 
 @RunWith(MockitoJUnitRunner.class)
 public class PathToPathMetadataTest {
 
   private static final String DIR_PATH = "file:/abc";
   private static final String FILE_PATH = DIR_PATH + "/child.dat";
-  private static final long LAST_MODIFIED_DIR = 123456789L;
+  private static final long LAST_MODIFIED_DIR = 0;
```
massdosage: This isn't sufficient to test your change, you probably need to keep this returning some non-zero value but in the assertions testing a folder below check that the last modification time is zero.

PR #53, opened 2018-03-21
Issue 52: Fail circus train if copy fails on tableReplicationStart

courtsVII:
Also relates to: https://github.com/HotelsDotCom/circus-train-bigquery/issues/5

### DIFF
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/CompositeTableReplicationListener.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/CompositeTableReplicationListener.java
@@ -18,6 +18,7 @@ package com.hotels.bdp.circustrain.core.event;
 import java.util.Collections;
 import java.util.List;
 
+import com.hotels.bdp.circustrain.api.CircusTrainException;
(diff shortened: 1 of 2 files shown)


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/CompositeTableReplicationListener.java
```diff
@@ -40,7 +41,7 @@ public void tableReplicationStart(EventTableReplication tableReplication, String
       try {
         listener.tableReplicationStart(tableReplication, eventId);
       } catch (Exception e) {
-        LOG.error("Listener '{}' threw exception on tableReplicationStart.", listener, e);
+        throw new CircusTrainException("Listener '" + listener + "' threw exception on tableReplicationStart.", e);
```
massdosage: Need a unit test for this code path.

patduin: Not sure if this is a good idea, cause it means the other listeners aren't run anymore.

See the other listener messages, they all log and we should end of in a state of replication failure.

PR #59, opened 2018-04-17
Using s3proxy with shaded guava

patduin:
* Using s3proxy with shaded jclouds and guava
* Removed hive-exec (the uberjar) in favor of hive-exec-core (the non uberjar with just the hive classes)
* lowered guava so hadoop can use it again.

s3Proxy dep blocked by new s3Proxy release, PR with changes is here: https://github.com/gaul/s3proxy/pull/273

Long story short, some users reported failing integration tests after some digging it looks like we had a classpath ordering problem with guava and either guava-20.0 or guava packaged in hive-exec.jar was used. After sorting that out I got consistent errors either due to s3proxy/jcloud needing guava 18.0 or higher or hadoop-common needed guava 17.0 or lower. Since we can't have to version on the classpath we need to shade which is was the PR for s3proxy tries to do. 

UPDATE:
Went for different approach where we do the shading ourselves.


pom.xml
```diff
@@ -62,7 +63,7 @@
     <beeju.version>1.2.1</beeju.version>
     <hamcrest-library.version>1.3</hamcrest-library.version>
     <housekeeping.version>1.0.2</housekeeping.version>
```
courtsVII: So where is Guava 20.0 being pulled in for Google Cloud?

patduin: I've downgraded guava all tests run? are we going to need guava-17 for tests and guava-20 for actual deployment?

patduin: @courtsVII so went back to 20.0 and override that version just in the test. I think we should be fine now as when we package it all up we relocate guava ourselves.

courtsVII: Looks good, approving

patduin:
On hold waiting for s3proxy release

circus-train-api/pom.xml
```diff
@@ -61,6 +61,7 @@
     <dependency>
       <groupId>org.apache.hive</groupId>
       <artifactId>hive-exec</artifactId>
+      <classifier>core</classifier>
```
massdosage: Why do we need to depend on the "core" classifier now?

patduin: hive-exec (is an uberjar)
hive-exec-core is the actual hive-exec hive code. We should if we can never rely on the non-core jar as it pulls in loads of stuff, guava among others.

massdosage: Agreed, makes sense.

massdosage:
What is the failure? If its something to do with the metrics then this is a known issue and probably unrelated to these changes.

patduin:
fixes  #60

PR #62, opened 2018-04-24
[issue-46] Replace SSH tunneling classes with library

### LINKED ISSUE #46: Extract SSH logic into a separate library
There is some shared SSH-tunnel code here and in Waggle Dance which needs to be improved. We want to pull the code out to a separate library.

ddcprg:
Fixes #46


circus-train-api/pom.xml
```diff
@@ -35,6 +35,12 @@
       <artifactId>metrics-core</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>com.hotels</groupId>
+      <artifactId>hcommon-ssh</artifactId>
+      <scope>provided</scope>
```
massdosage: Why is the scope for this "provided"? This is saying we expect someone else to have put it on the classpath?

ddcprg: this is an error, probably copy&pasta

ddcprg: Actually this doesn't matter because is the API, `circus-train-core` provides the library

massdosage: OK, so can you remove it?

ddcprg: no, I need the marker for the closeable metastore interface

massdosage: So what does "provided" get you here as opposed to not listing a scope at all?

ddcprg: make it not transitive?

massdosage: Yeah, but leaving the scope out would do the same. "provided" you usually use if someone else is going to put the jars on the classpath (as we often do with Hadoop being provided). I just don't understand what "provided" gets us here?

ddcprg: the default is compile which makes the dependency transitive, i.e. any downstream implementation of the API will pull the SSH library which they probably don't need. This is all theoretical

massdosage: Yeah, we don't give any other dependency here this special treatment which is why this one stands out. I can live with it I guess.

PR #67, opened 2018-05-01
Fix tests

ddcprg:
I'm still relying on Travis for this, I can't compile locally

### DIFF
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/MetastoreTunnel.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/MetastoreTunnel.java
@@ -96,7 +96,7 @@ public class MetastoreTunnel {
   }
 
   public void setStrictHostKeyChecking(String strictHostKeyChecking) {
-    this.strictHostKeyChecking = strictHostKeyChecking == null ? null : strictHostKeyChecking.toLowerCase();
+    this.strictHostKeyChecking = strictHostKeyChecking;
   }
 
 }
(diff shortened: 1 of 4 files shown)


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/metastore/TunnellingMetaStoreClientSupplier.java
```diff
@@ -66,15 +66,16 @@ private static int getLocalPort() {
...
     private final MetaStoreClientFactory metaStoreClientFactory;
-    private final HiveConf hiveConf;
     private final String name;
+    final @VisibleForTesting HiveConf hiveConf;
 
     private HiveMetaStoreClientSupplier(MetaStoreClientFactory metaStoreClientFactory, HiveConf hiveConf, String name) {
       this.metaStoreClientFactory = metaStoreClientFactory;
-      this.hiveConf = hiveConf;
       this.name = name;
+      this.hiveConf = hiveConf;
```
massdosage: Why was this moved? The order before matched the order the arguments are passed in to the constructor.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/MetastoreTunnel.java
```diff
@@ -96,7 +96,7 @@ public String getStrictHostKeyChecking() {
   }
 
   public void setStrictHostKeyChecking(String strictHostKeyChecking) {
-    this.strictHostKeyChecking = strictHostKeyChecking == null ? null : strictHostKeyChecking.toLowerCase();
+    this.strictHostKeyChecking = strictHostKeyChecking;
```
massdosage: So you don't need to do the null check and lower casing because this is handled somewhere else?

PR #75, opened 2018-06-06
Avro schema replication failing when avro.schema.url without scheme is specified

### LINKED ISSUE #74: Avro schema replication failing when avro.schema.url without scheme is specified

courtsVII:
fixes #74


circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformation.java
```diff
@@ -50,6 +55,28 @@ protected boolean avroTransformationSpecified() {
...
+    if (overrideScheme != null && StringUtils.isNotBlank(overrideScheme.toString())) {
+      scheme = overrideScheme.toString();
+    } else {
+      scheme = avroSerDeConfig.getSourceAvroUrlScheme();
+    }
+
+    if (isNotEmpty(scheme)) {
+      try {
+        URI uri = new URI(source);
+        source = new URI(scheme, uri.getUserInfo(), uri.getHost(), uri.getPort(), uri.getPath(), uri.getQuery(),
```
massdosage: Would all this work if instead of doing all this you just prefixed "source" with the passed in scheme and did nothing more? i.e. none of this URI stuff, you just append the two strings and pass it back out?

circus-train-avro/src/test/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformationTest.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  @Test
+  public void constructSourceOverridesScheme() {
+    AvroSerDeConfig config = new AvroSerDeConfig();
+    config.setSourceAvroUrlScheme("hdfs");
+    AbstractAvroSerDeTransformation transformation = new AvroSerDePartitionTransformation(config, null);
+    assertEquals("hdfs:/etl/EAN/avsc/tlog_throttled_event.avsc", transformation.constructSource("s3:///etl/EAN/avsc/tlog_throttled_event.avsc"));
+  }
+
```
massdosage: Could do with some more edge cases - these come to mind
```
config.setSourceAvroUrlScheme("");
config.setSourceAvroUrlScheme("   ");
config.setSourceAvroUrlScheme("someinvalid:uri://scheme?");
```

circus-train-avro/src/test/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformationTest.java
```diff
@@ -0,0 +1,48 @@
...
+    AbstractAvroSerDeTransformation transformation = new AvroSerDePartitionTransformation(config, null);
+    assertEquals("hdfs:/etl/EAN/avsc/tlog_throttled_event.avsc", transformation.constructSource("/etl/EAN/avsc/tlog_throttled_event.avsc"));
+  }
+
+  @Test
+  public void constructSourceOverridesScheme() {
+    AvroSerDeConfig config = new AvroSerDeConfig();
+    config.setSourceAvroUrlScheme("hdfs");
+    AbstractAvroSerDeTransformation transformation = new AvroSerDePartitionTransformation(config, null);
+    assertEquals("hdfs:/etl/EAN/avsc/tlog_throttled_event.avsc", transformation.constructSource("s3:///etl/EAN/avsc/tlog_throttled_event.avsc"));
```
massdosage: I'm not sure you'd ever want to do this (replace e.g. s3: with hdfs:) but this would be an argument for reconstructing the URI rather than just appending the strings.

circus-train-avro/src/test/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformationTest.java
```diff
@@ -0,0 +1,48 @@
...
+import com.hotels.bdp.circustrain.avro.conf.AvroSerDeConfig;
+
+public class AbstractAvroSerDeTransformationTest {
+
+  @Test
+  public void constructSourceWithScheme() {
+    AvroSerDeConfig config = new AvroSerDeConfig();
+    config.setSourceAvroUrlScheme("hdfs");
+    AbstractAvroSerDeTransformation transformation = new AvroSerDePartitionTransformation(config, null);
+    assertEquals("hdfs:/etl/EAN/avsc/tlog_throttled_event.avsc", transformation.constructSource("/etl/EAN/avsc/tlog_throttled_event.avsc"));
```
massdosage: Is hdfs:/ OK or should it be hdfs:/// ?

courtsVII: hdfs:/ resolves to hfs:///

circus-train-avro/src/test/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformationTest.java
courtsVII: Yeah I don't think this is recommended however its more consistent, the alternative is silently chucking away what the user configured if there is already a scheme on their URL (or we could just throw an exception)

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformation.java
courtsVII: Lots of edge cases, this is better imo

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformation.java
```diff
@@ -50,6 +55,28 @@ protected boolean avroTransformationSpecified() {
...
+    String scheme = null;
+    Object overrideScheme = avroSerdeConfigOverride
+        .get(AvroSerDeConfig.TABLE_REPLICATION_OVERRIDE_SOURCE_AVRO_URL_SCHEME);
+    if (overrideScheme != null && StringUtils.isNotBlank(overrideScheme.toString())) {
+      scheme = overrideScheme.toString();
+    } else {
+      scheme = avroSerDeConfig.getSourceAvroUrlScheme();
+    }
+
+    if (isNotEmpty(scheme)) {
```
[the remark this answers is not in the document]
courtsVII: It doesn't return null it returns the original String with no changes

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformation.java
```diff
@@ -50,6 +55,28 @@ protected boolean avroTransformationSpecified() {
     return argsPresent(getAvroSchemaDestinationFolder(), getEventId());
   }
 
+  protected String constructSource(String source) {
```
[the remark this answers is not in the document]
courtsVII: It won't get this far if the schema url isn't being overriden

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/conf/AvroSerDeConfig.java
```diff
@@ -33,4 +36,12 @@ public String getBaseUrl() {
   public void setBaseUrl(String avroSchemaBaseUrl) {
     this.avroSchemaBaseUrl = avroSchemaBaseUrl;
   }
+
+  public String getSourceAvroUrlScheme() {
```
patduin: Can you not get the scheme from the configured hiveConf? path.getFileSystem(sourceHiveConf).getScheme()?

circus-train-avro/src/test/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformationTest.java
```diff
@@ -0,0 +1,76 @@
...
+import com.hotels.bdp.circustrain.api.CircusTrainException;
+import com.hotels.bdp.circustrain.avro.conf.AvroSerDeConfig;
+
+public class AbstractAvroSerDeTransformationTest {
+
+  @Test
+  public void constructSourceWithScheme() {
+    AvroSerDeConfig config = new AvroSerDeConfig();
+    config.setSourceAvroUrlScheme("hdfs");
+    AbstractAvroSerDeTransformation transformation = new AvroSerDePartitionTransformation(config, null);
```
patduin: These are `AvroSerDePartitionTransformation` Tests, move them there and add them to the other concrete classes as well. Abstract class don't really have tests, unless you copy the hierarchy in the test class, which is not great.
This is where composition over inheritance comes in.

circus-train-avro/src/test/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformationTest.java
```diff
@@ -0,0 +1,76 @@
...
+    assertEquals("hdfs:/etl/EAN/avsc/tlog_throttled_event.avsc", transformation.constructSource("/etl/EAN/avsc/tlog_throttled_event.avsc"));
+  }
+
+  @Test
+  public void constructSourceOverridesScheme() {
+    AvroSerDeConfig config = new AvroSerDeConfig();
+    config.setSourceAvroUrlScheme("hdfs");
+    AbstractAvroSerDeTransformation transformation = new AvroSerDePartitionTransformation(config, null);
+    assertEquals("hdfs:/etl/EAN/avsc/tlog_throttled_event.avsc", transformation.constructSource("s3:///etl/EAN/avsc/tlog_throttled_event.avsc"));
```
patduin: assertThat(...,  is(expected)) please.

circus-train-avro/src/test/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformationTest.java
patduin: I think we should get it from the FileSystem then this becomes a non-issue

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/conf/AvroSerDeConfig.java
courtsVII: I'm not sure if this is a good idea with how we transform the FileSystem in Circus Train, getScheme is implementation specific and change the FileSystem depending on where the source and destination of the replication resides. Here is the getScheme methods for DistributedFileSystem, S3FileSystem and GoogleHadoopFileSystem respectively:

``` java
public String getScheme() {
    return "hdfs";
}

public String getScheme() {
    return "s3";
}

public String getScheme() {
    return "gs";
}
```

patduin: That is my point depending on the config we get the correct FileSystem and thus the correct scheme. Why is the example you give wrong?

courtsVII: I'm not sure that's the case with how we're binding the filesystems, for example say your source is in Chandler and destination is in S3 we would bind the S3FileSystem. In Chandler an avro.schema.url with no scheme will default to HDFS, but if we bind the S3FileSystem in an initialization step which I believe we do then the default scheme becomes S3 and the file won't be found. I will test it I'm not 100% sure on the ordering of things and when certain things are triggered

patduin: Please check. My understanding is that you still get a DistributedFileSystemt back. 
The binding does: fs.s3.impl=foo.bar.S3FileSystem.class. Which means if you need to load a S3 FileSystem use this particular implementation. It doesn't say if you create a fileSystem always create an S3FileSystem.
Do you really get an S3FileSystem back?

patduin: Hmm it might fall back to "file" scheme (unless this is set: 'fs.defaultFS'), would be interesting to see what you find out.

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/util/FileSystemPathResolver.java
```diff
@@ -17,24 +17,49 @@
...
-  public Path resolve(String url) {
+  public Path resolveScheme(Path path) {
+    return resolveScheme(path.toString());
+  }
+
+  public Path resolveScheme(String url) {
+    try {
+      URI uri = new URI(url);
+      if (isEmpty(uri.getScheme())) {
+        String scheme = FileSystem.get(configuration).getScheme();
```
massdosage: OK, so this is saying that if the scheme is empty, we get a default scheme from the config and use that? And for the setup that raised this issue this turns out to be "hdfs" so it resolves the issue? If so then the comment in the changelog about the new config param needs to be removed as there isn't one anymore.

circus-train-avro/src/test/java/com/hotels/bdp/circustrain/avro/util/FileSystemPathResolverTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.junit.Test;
+
+public class FileSystemPathResolverTest {
+
+  private Configuration configuration = new Configuration();
+
+  @Test
+  public void resolveSchemeDoesntChangeSchemeForPathWithScheme() {
+    Path path = new Path("s3:/etl/test/avsc/schema.avsc");
+    assertEquals(path, new FileSystemPathResolver(configuration).resolveScheme(path));
```
patduin: I think in general we prefer hamcrest: `assertThat(new FileSystemPathResolver(configuration).resolveScheme(path), is(path));` As you get nicer error reporting

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 ## TBD
+### Added
+* A configuration parameter to add or replace a scheme on your source avro.schema.url. See [#74](https://github.com/HotelsDotCom/circus-train/issues/74)
```
patduin: yeah as Adrian said this can go, but I understand you're still testing a bit.

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/util/FileSystemPathResolver.java
```diff
@@ -17,24 +17,51 @@
...
   }
 
-  public Path resolve(String url) {
+  public Path resolveScheme(Path path) {
+    return resolveScheme(path.toString());
+  }
+
+  public Path resolveScheme(String url) {
+    try {
+      URI uri = new URI(url);
```
massdosage: Could we make the "main" way of calling this functionality go via a Path instead of a String and then to Path.toUri()? Would be "safer" than this and probably no need for try/catch.

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/util/SchemaCopier.java
```diff
@@ -56,15 +56,16 @@ public Path copy(String source, String destination) {
 
     java.nio.file.Path temporaryDirectory = createTempDirectory();
 
-    Path sourceLocation = new NameServicePathResolver(sourceHiveConf).resolve(source);
+    FileSystemPathResolver sourceFileSystemPathResolver = new FileSystemPathResolver(sourceHiveConf);
+    Path sourceLocation = sourceFileSystemPathResolver
+        .resolveNameServices(sourceFileSystemPathResolver.resolveScheme(source));
```
massdosage: Split this across more lines, will be easier to read

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/util/FileSystemPathResolver.java
```diff
@@ -0,0 +1,80 @@
...
+    try {
+      URI uri = path.toUri();
+      if (isEmpty(uri.getScheme())) {
+        String scheme = FileSystem.get(configuration).getScheme();
+        Path result = new Path(new URI(scheme, uri.getUserInfo(), uri.getHost(), uri.getPort(), uri.getPath(),
+            uri.getQuery(), uri.getFragment()).toString());
+        LOG.info("Added scheme {} to path {}. Resulting path is {}", scheme, path, result);
+        return result;
+      }
+    } catch (URISyntaxException | IOException e) {
```
massdosage: Do we need either/both of these? If neither then we can probably get rid of the whole try/catch? Or just remove one of the exception types?

courtsVII: URISyntaxException is needed for new URI(....) with the new scheme and IOException for getting the FileSystem

massdosage: Ah yes, of course, the new URI() is where it comes from, makes sense.

PR #83, opened 2018-07-31
Load Google Cloud credentials from Absolute or Relative path

### LINKED ISSUE #82: Configurable working directory for GCP keys
### Background
It was noted that the current Circus Train BigQuery plugin implementation attempts to create a copy of the GCP credentials JSON file in the current working directory in order to configure DistCp jobs properly.

### Issue
The replication flow fails as the resolved working directory is owned by _root_ and the replication script is executed by the user _hadoop_.
```
Caused by: java.io.FileNotFoundException: /foo/bar/azkaban-executor-2.5.0/executions/94082/users-key.json (Permission denied)
```

AnanaMJ:
Fixes #82 

Loading from relative path allows us to skip copy to `user.dir` step which was failing Azkaban jobs.


circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -32,60 +34,89 @@
...
 import com.hotels.bdp.circustrain.api.CircusTrainException;
+import com.hotels.bdp.circustrain.gcp.context.GCPSecurity;
 
 class GCPCredentialCopier {
   private static final Logger LOG = LoggerFactory.getLogger(GCPCredentialCopier.class);
-  private static final String WORKING_DIRECTORY = System.getProperty("user.dir");
+
+  private static final String DEFAULT_WORKING_DIRECTORY = System.getProperty("user.dir");
+  private static final String DEFAULT_HDFS_PREFIX = "hdfs:///tmp/ct-gcp-";
   private static final String RANDOM_STRING = UUID.randomUUID().toString() + System.currentTimeMillis();
```
teabot: This is set only once, when the class is loaded. Is this what we require?

teabot: Seems to be a lot going on here regarding the construction of paths. Perhaps better to move into a mockable `GCPCredentialPathFactory` or similar?

teabot: Its nice to pass in a mockable `Clock` so that tests can be independent of the time domain.

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -32,60 +34,89 @@
...
+    LOG.debug("Credential Provider URI = {}", credentialProvider);
+
+    hdfsGsCredentialDirectory = getHdfsGsCredentialDirectory();
+    hdfsGsCredentialAbsolutePath = hdfsGsCredentialDirectory + CACHED_CREDENTIAL_NAME;
+    LOG.debug("Temporary HDFS Google Cloud credential location set to {}", hdfsGsCredentialDirectory);
+    LOG.debug("HDFS Google Cloud credential path will be {}", hdfsGsCredentialAbsolutePath);
+  }
+
+  private String getHdfsGsCredentialDirectory() {
+    return isBlank(security.getDistributedFileSystemWorkingDirectory()) ? DEFAULT_HDFS_PREFIX + RANDOM_STRING
```
teabot: ternary op is not favoured in these projects IIRC.

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -32,60 +34,89 @@
...
   }
 
   void copyCredentials() {
     try {
-      copyCredentialIntoWorkingDirectory();
-      copyCredentialIntoHdfs();
-      copyCredentialIntoDistributedCache();
+      if (new File(credentialProvider).isAbsolute()) {
+        LOG.debug("Copying Google Cloud credentials from absolute path {}", credentialProvider);
+        copyCredentialIntoWorkingDirectory();
```
teabot: Why copying into 3 places? Will not one suffice? A comment here may help.

teabot: These different copy operations feel like they might better exist as 3 separate implementations (classes) with a common abstraction? However, I accept that this is beyond the scope of this issue.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialConfigurerTest.java
```diff
@@ -43,7 +44,7 @@
 @PrepareForTest({ FileSystem.class })
 public class GCPCredentialConfigurerTest {
 
-  private @Rule TemporaryFolder temporaryFolder = new TemporaryFolder();
+  private @Rule final TemporaryFolder temporaryFolder = new TemporaryFolder();
   private @Mock FileSystem mockFS;
```
teabot: I usual prefer the omission of the `mock` prefix, and no abbreviation of the variable name: `FS` → `FileSystem`.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -15,60 +15,159 @@
...
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.times;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.verify;
 import static org.powermock.api.mockito.PowerMockito.doNothing;
 import static org.powermock.api.mockito.PowerMockito.doReturn;
+import static org.powermock.api.mockito.PowerMockito.mockStatic;
+import static org.powermock.api.mockito.PowerMockito.verifyStatic;
```
teabot: PowerMock lea͠ki̧n͘g fr̶ǫm ̡m​͟y​͟ eye͢s̸ ̛l̕ik͏e liq​uid pain :-)

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -32,60 +34,89 @@
...
+    return isBlank(security.getDistributedFileSystemWorkingDirectory()) ? DEFAULT_HDFS_PREFIX + RANDOM_STRING
+        : security.getDistributedFileSystemWorkingDirectory();
   }
 
   void copyCredentials() {
     try {
-      copyCredentialIntoWorkingDirectory();
-      copyCredentialIntoHdfs();
-      copyCredentialIntoDistributedCache();
+      if (new File(credentialProvider).isAbsolute()) {
```
teabot: What is in `credentialProvider`, name could be more revealing of contents.

teabot: might be better to pull the `File` instantiation into a local var for easier debugging.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -16,59 +16,128 @@
...
 import org.mockito.Mock;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 
 import fm.last.commons.test.file.TemporaryFolder;
 
 import com.hotels.bdp.circustrain.api.CircusTrainException;
+import com.hotels.bdp.circustrain.gcp.context.GCPSecurity;
 
 @RunWith(PowerMockRunner.class)
```
patduin: change to mockito runner

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -15,77 +15,107 @@
...
+  private String getHdfsGsCredentialDirectory() {
+    if (isBlank(security.getDistributedFileSystemWorkingDirectory())) {
+      return DEFAULT_HDFS_PREFIX + RANDOM_STRING;
+    } else {
+      return security.getDistributedFileSystemWorkingDirectory();
     }
-    this.credentialProvider = credentialProvider;
+  }
+
+  private String getCredentialProviderRelativePath() {
```
courtsVII: You can add 
```if (new File(security.getCredentialProvider).isAbsolute()) {
           //relativize
     } else {
           //return unmodified path
     }
```
So that users can configure it either way rather than job failing if they provide an absolute path in the config

AnanaMJ: That's a great idea, I'll add this now.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -16,59 +16,128 @@
...
+  @Test
+  public void shouldGetRelativePath() throws IOException {
+    temporaryFolder.newFile(credentialsFilePath);
+    setGcpSecurity(credentialsFilePath, null);
+    GCPCredentialCopier copier = new GCPCredentialCopier(fileSystem, conf, gcpSecurity);
+    ArgumentCaptor<Path> pathCaptor = ArgumentCaptor.forClass(Path.class);
     copier.copyCredentials();
+    verify(fileSystem).copyFromLocalFile(pathCaptor.capture(), any(Path.class));
+    Path source = pathCaptor.getValue();
+    assertTrue(source.toString().equals(credentialsFileRelativePath));
```
patduin: This give you very bad error messages if it fails prefer: `assertThat(source.toString(), is(credentialsFileRelativePath));`

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -16,59 +16,128 @@
...
   }
 
   @Test
-  public void copyCredentials() throws Exception {
-    doNothing().when(fs).copyFromLocalFile(any(Path.class), any(Path.class));
-    String credentialProvider = "test.json";
-    temporaryFolder.newFile(credentialProvider);
-    credentialProvider = temporaryFolder.getRoot() + "/" + credentialProvider;
-    GCPCredentialCopier copier = new GCPCredentialCopier(fs, conf, credentialProvider);
+  public void copyCredentialsWithCredentialProviderSupplied() throws Exception {
```
patduin: Remove the `shouldGetRelativePath` test method and combine the assert you do there in one test like this:


```
 @Test
  public void copyCredentialsWithCredentialProviderSupplied() throws Exception {
    setGcpSecurity(credentialsFilePath, null);
    GCPCredentialCopier copier = new GCPCredentialCopier(fileSystem, conf, gcpSecurity);
    copier.copyCredentials();

    verify(fileSystem).copyFromLocalFile(eq(new Path(credentialsFileRelativePath)), any(Path.class));
    assertNotNull(conf.get(DISTRIBUTED_CACHE_PROPERTY));
    assertThat(conf.get(DISTRIBUTED_CACHE_PROPERTY), endsWith(SYMLINK_FLAG + credentialsFileRelativePath));
  }
  }

```

imo, it more clear that you copy does two things, copy to hdfs and make the cache entry.

Then add another test where the credentialsFilePath is configured as a relative file path.

A further improvement would be like @teabot  suggested to add a factory for the `hdfsGsCredential` paths. E.g. `HdfsGsCredentialPathGenerator`, has one method `newPath()`.  You can add the random stuff in there and mock the class in this test to do better verifications.
You can't assert the deleteOnExit very well atm, this generator class will help with that.

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -15,77 +15,107 @@
...
-    Path destination = new Path(HDFS_GS_CREDENTIAL_ABSOLUTE_PATH);
-    Path destinationFolder = new Path(HDFS_GS_CREDENTIAL_DIRECTORY);
+    /*
+     * The Google credentials file must be present in HDFS so that the DistCP map reduce job can access it upon
+     * replication.
+     */
+    Path source = new Path(credentialsFileRelativePath);
+    Path destination = new Path(hdfsGsCredentialAbsolutePath);
+    Path destinationFolder = new Path(hdfsGsCredentialDirectory);
     fs.deleteOnExit(destinationFolder);
```
patduin: This is not tested, assert it in the test I suggested.

The destinationFolder looks wrong, it could be the working directory that a user configures `security.getDistributedFileSystemWorkingDirectory()`, which you will delete. What if this working directory is used by other jobs you'll be deleting their data. 
Only delete folders/files you create.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -16,59 +16,128 @@
...
+@PrepareForTest({ FileSystem.class, FileUtils.class })
 public class GCPCredentialCopierTest {
 
-  private @Rule TemporaryFolder temporaryFolder = new TemporaryFolder();
-  private @Mock FileSystem fs;
-  private Configuration conf = new Configuration();
+  private static final String DISTRIBUTED_CACHE_PROPERTY = "mapreduce.job.cache.files";
+  private static final String SYMLINK_FLAG = "#";
+
+  private @Rule final TemporaryFolder temporaryFolder = new TemporaryFolder();
```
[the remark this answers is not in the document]
patduin: I suspect you don't actually need the rule as you're not really moving files around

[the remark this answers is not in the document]
AnanaMJ: Seems like it is just powerMockRunner making it work, but with MockitoJUnitRunner, it needs to be public. If I change it to `public @Mock`, it works.

I see now why @teabot added the comment about PowerMock 😂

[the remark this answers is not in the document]
patduin: `public @Rule` I think you mean. But yeah you might be able to remove it altogether you don't really need to create files for this test.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
AnanaMJ: I added factories for building some paths, so tests can be more separate from each other and some elements can be mocked, which makes testing a bit easier. I'm going to push the changes as soon as I finish testing the build.

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/HdfsGsCredentialDirectoryFactory.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.hadoop.fs.Path;
+
+import com.hotels.bdp.circustrain.gcp.context.GCPSecurity;
+
+class HdfsGsCredentialDirectoryFactory {
+
+  static final String DEFAULT_HDFS_PREFIX = "hdfs:/tmp/ct-gcp-";
+  private String hdfsBaseDirectory;
+
+  Path newInstance(GCPSecurity security, String randomString) {
```
patduin: I think you still might have issues here.
The default is fine, but you're changing the working dir.
Example:
```security.getDistributedFileSystemWorkingDirectory() == /home/user/```

Now you're code is going to change that to ```/home/user-random/```
Which means you need to have write permissions in `/home/` which you probably don't.

Make a random subdir it will solve everything.

`return new Path(hdfsBaseDirectory, randomString);`
Notice the ',' instead of +

This directory will be save to delete as well.

File: circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/HdfsGsCredentialAbsolutePathFactory.java [hunk 14]
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.gcp;
+
+import org.apache.hadoop.fs.Path;
+
+public class HdfsGsCredentialAbsolutePathFactory {
+
+  Path newInstance(Path directory, String randomString) {
```
patduin: Since you're making the dir random, you don't need to make the file random as well, so the random string can go, which makes this class trivial and it can go altogether, just inline it in the copier class. One factory less.

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/HdfsGsCredentialAbsolutePathFactory.java
```diff
@@ -0,0 +1,27 @@
...
+    String absolutePath = directory + "/ct-gcp-key-" + randomString + ".json";
+    return new Path(absolutePath);
```
patduin: In general have a little look at the Path api. It has constructors new Path(parent, child) which is the preferred way of creating path. Preferred over string manipulation, handles the separator (/) etc...

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -15,77 +15,105 @@
...
+
+    if (security == null || isBlank(security.getCredentialProvider())) {
+      throw new IllegalArgumentException("gcp-security credential-provider must be set");
     }
-    this.credentialProvider = credentialProvider;
+    this.security = security;
+
+    randomString = this.randomStringFactory.newInstance();
+    hdfsGsCredentialDirectory = this.hdfsGsCredentialDirectoryFactory.newInstance(security, randomString);
+    hdfsGsCredentialAbsolutePath = this.hdfsGsCredentialAbsolutePathFactory
```
patduin: See my other comment you probably don't need a factory for this.

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -15,77 +15,105 @@
...
+  private final HdfsGsCredentialDirectoryFactory hdfsGsCredentialDirectoryFactory;
+  private final RandomStringFactory randomStringFactory;
+  private final String credentialsFileRelativePath;
+  private final String randomString;
+  private final Path hdfsGsCredentialDirectory;
+  private final Path hdfsGsCredentialAbsolutePath;
+
+  GCPCredentialCopier(FileSystem fs, Configuration conf, GCPSecurity security) {
+    this(fs, conf, security, new CredentialProviderRelativePathFactory(), new HdfsGsCredentialAbsolutePathFactory(),
+        new HdfsGsCredentialDirectoryFactory(), new RandomStringFactory());
```
patduin: Remove the `RandomStringFactory` but add it to the constructor of the Factories that need it. 
example: `new CredentialProviderRelativePathFactory(new RandomStringFactory())`

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/CredentialProviderRelativePathFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.gcp;
+
+import java.nio.file.Paths;
+
+import com.hotels.bdp.circustrain.gcp.context.GCPSecurity;
+
```
courtsVII: This class could probably have a better name, maybe simply `GCPCredentialProvider` the factory is implementation details, then you could just pass the `GCPCredentialProvider` to the `GCPCredentialCopier` and `@Autowire` this class using it as a replacement for the GCPSecurity POJO, and do similar for the `HdfsGsCredentialAbsolutePathFactory`

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -35,57 +31,60 @@
 
 class GCPCredentialCopier {
   private static final Logger LOG = LoggerFactory.getLogger(GCPCredentialCopier.class);
-  private static final String WORKING_DIRECTORY = System.getProperty("user.dir");
-  private static final String RANDOM_STRING = UUID.randomUUID().toString() + System.currentTimeMillis();
-  private static final String CACHED_CREDENTIAL_NAME = "/ct-gcp-key-" + RANDOM_STRING + ".json";
-  private static final String HDFS_GS_CREDENTIAL_DIRECTORY = "hdfs:///tmp/ct-gcp-" + RANDOM_STRING;
-  private static final String HDFS_GS_CREDENTIAL_ABSOLUTE_PATH = HDFS_GS_CREDENTIAL_DIRECTORY + CACHED_CREDENTIAL_NAME;
+  static final String GCP_KEY_NAME = "ct-gcp-key.json";
```
courtsVII: You probably still want this to have a random String on the end in case user configures a job to copy credentials to specific HDFS directory and it runs more than once. Consider moving the logic `dfsGCPCredentialAbsolutePath = new Path(dfsGCPCredentialDirectory, GCP_KEY_NAME)` into `DistributedFileSystemPathProvider` and append random String to the file there, return absolute path to credential file from `newPath` and use `path.getParent()` to get the `dfsGCPCredentialDirectory`

courtsVII: Ah I see you add random string to a subdirectory in `DistributedFileSystemPathProvider`. Still move appending the key to the Path into that class so the logic is centralized

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/DistributedFileSystemPathProvider.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.hadoop.fs.Path;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import com.hotels.bdp.circustrain.gcp.context.GCPSecurity;
+
+@Component
+public class DistributedFileSystemPathProvider {
+
+  static final String DEFAULT_HDFS_PREFIX = "hdfs:/tmp/ct-gcp-";
```
patduin: Sorry I should have noticed this earlier but instead of `tmp` you should probable use the value of `hive.exec.scratchdir` from the HiveConf. See https://cwiki.apache.org/confluence/display/Hive/AdminManual+Configuration#AdminManualConfiguration-TemporaryFolders

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -15,60 +15,67 @@
...
-  public void credentialProviderNotSetThrowsException() throws Exception {
-    doNothing().when(fs).copyFromLocalFile(any(Path.class), any(Path.class));
-    doReturn(false).when(fs).exists(any(Path.class));
-    String credentialProvider = null;
-    GCPCredentialCopier copier = new GCPCredentialCopier(fs, conf, credentialProvider);
+  @Before
+  public void init() {
+    doReturn(credentialsFileRelativePath).when(credentialPathProvider).newPath();
+    doReturn(dfsAbsolutePath).when(distributedFileSystemPathProvider).newPath();
+    // doReturn(dfsDirectory).when(distributedFileSystemPathProvider).newPath().getParent();
```
patduin: remove

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -15,60 +15,67 @@
...
-  private @Rule TemporaryFolder temporaryFolder = new TemporaryFolder();
-  private @Mock FileSystem fs;
-  private Configuration conf = new Configuration();
+  private static final String DISTRIBUTED_CACHE_PROPERTY = "mapreduce.job.cache.files";
+  private static final String SYMLINK_FLAG = "#";
+
+  private final Configuration conf = new Configuration();
+  private final Path credentialsFileRelativePath = new Path("../test.json");
+  private final Path dfsDirectory = new Path("/rootDirectory/test-random-string");
+  private final Path dfsAbsolutePath = new Path(dfsDirectory + "/" + DistributedFileSystemPathProvider.GCP_KEY_NAME);
```
patduin: `new Path(dfsDirectory, DistributedFileSystemPathProvider.GCP_KEY_NAME)`

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialPathProviderTest.java
```diff
@@ -0,0 +1,60 @@
...
+    String result = new GCPCredentialPathProvider(security).newPath().toString();
+    assertThat(result, is(relativePath));
+  }
+
+  @Test
+  public void newInstanceWithAbsolutePath() {
+    security.setCredentialProvider("/test.json");
+    String result = new GCPCredentialPathProvider(security).newPath().toString();
+    assertFalse(new Path(result).isAbsolute());
+    assertTrue(result.startsWith("../"));
```
patduin: `assertThat(result, startsWith("../"));` please, gives much nicer fail message

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopier.java
```diff
@@ -17,75 +17,59 @@
...
-    conf.set(GCP_KEYFILE_CACHED_LOCATION, "." + CACHED_CREDENTIAL_NAME);
+  private void linkRelativePathInDistributedCache(Configuration conf, Path source, Path destination)
+    throws URISyntaxException, IOException {
+    /*
+     * The "#" links the HDFS location for the key file to the local file system credential provider path so that the
+     * GoogleHadoopFileSystem can subsequently resolve it from a local file system uri despite it being in a Distributed
+     * file system when the DistCP job runs.
+     */
+    String cacheFileUri = destination.toString() + "#" + source;
+    org.apache.hadoop.mapreduce.filecache.DistributedCache.addCacheFile(new URI(cacheFileUri), conf);
```
patduin: can we import the `DistributedCache`?

circus-train-gcp/src/main/java/com/hotels/bdp/circustrain/gcp/DistributedFileSystemPathProvider.java
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      Path parentDirectory = new Path(security.getDistributedFileSystemWorkingDirectory(), randomString);
+      return new Path(parentDirectory, GCP_KEY_NAME);
+    }
+
+  }
+
+  private Path getTemporaryFolder(Configuration configuration, String randomString) {
+    String temporaryFolder = configuration.get("hive.exec.scratchdir");
+    if (isBlank(temporaryFolder)) {
```
courtsVII: Replace raw String with `DEFAULT_HDFS_PREFIX`

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -15,60 +15,66 @@
...
+  private final Configuration conf = new Configuration();
+  private final Path credentialsFileRelativePath = new Path("../test.json");
+  private final Path dfsDirectory = new Path("/rootDirectory/test-random-string");
+  private final Path dfsAbsolutePath = new Path(dfsDirectory, DistributedFileSystemPathProvider.GCP_KEY_NAME);
+
+  private @Mock FileSystem fileSystem;
+  private @Mock GCPCredentialPathProvider credentialPathProvider;
+  private @Mock DistributedFileSystemPathProvider distributedFileSystemPathProvider;
+
+  private final GCPCredentialCopier copier = new GCPCredentialCopier();
```
courtsVII: Move this in line with the other `private final` fields

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/context/GCPBeanPostProcessorTest.java
```diff
@@ -16,46 +16,78 @@
...
+  private @Mock Configuration configuration;
+  private @Mock BindGoogleHadoopFileSystem bindGoogleHadoopFileSystem;
+  private @Mock FileSystemFactory fileSystemFactory;
+  private @Mock GCPCredentialCopier credentialCopier;
+
+  private GCPBeanPostProcessor processor;
+
+  @Before
+  public void init() {
+    doNothing().when(configuration).set(anyString(), anyString());
```
patduin: remove this is the default behavour of mockito

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/context/GCPBeanPostProcessorTest.java
```diff
@@ -16,46 +16,78 @@
...
+    verify(credentialPathProvider).newPath();
   }
 
   @Test
   public void postProcessAfterInitializationWithIncorrectBeanName() throws Exception {
     String beanName = "notBaseConf";
-    GCPBeanPostProcessor processor = new GCPBeanPostProcessor(security);
     processor.postProcessAfterInitialization(configuration, beanName);
-    verify(security, times(0)).getCredentialProvider();
+    verify(credentialPathProvider, times(0)).newPath();
```
patduin: Replace with:  `verifyZeroInteractions(credentialPathProvider, bindGoogleHadoopFileSystem, credentialCopier);`

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/context/GCPBeanPostProcessorTest.java
```diff
@@ -16,46 +16,78 @@
...
+import com.hotels.bdp.circustrain.gcp.FileSystemFactory;
+import com.hotels.bdp.circustrain.gcp.GCPCredentialCopier;
+import com.hotels.bdp.circustrain.gcp.GCPCredentialPathProvider;
+import com.hotels.bdp.circustrain.gcp.RandomStringFactory;
 
 @RunWith(MockitoJUnitRunner.class)
 public class GCPBeanPostProcessorTest {
-  private @Spy GCPSecurity security = new GCPSecurity();
-  private @Spy Configuration configuration = new Configuration();
+  private @Mock RandomStringFactory randomStringFactory;
```
patduin: remove

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/context/GCPBeanPostProcessorTest.java
```diff
@@ -16,46 +16,78 @@
...
+
+  private @Mock GCPSecurity security;
```
patduin: remove

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/context/GCPBeanPostProcessorTest.java
```diff
@@ -16,46 +16,78 @@
...
   public void postProcessAfterInitializationWithIncorrectBeanName() throws Exception {
     String beanName = "notBaseConf";
-    GCPBeanPostProcessor processor = new GCPBeanPostProcessor(security);
     processor.postProcessAfterInitialization(configuration, beanName);
-    verify(security, times(0)).getCredentialProvider();
+    verify(credentialPathProvider, times(0)).newPath();
   }
 
   @Test
   public void postProcessAfterInitializationWithBlankCredentialProviderDoesntModifyConfiguration() throws Exception {
```
patduin: remove this test, your code doesn't modify the configuration so you should not have to test it, it's also using `security` which is probably a remnant of a previous refactoring.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/context/GCPBeanPostProcessorTest.java
```diff
@@ -16,46 +16,78 @@
...
+
+  @Before
+  public void init() {
+    doNothing().when(configuration).set(anyString(), anyString());
+    processor = new GCPBeanPostProcessor(credentialPathProvider, distributedFileSystemPathProvider,
+        bindGoogleHadoopFileSystem, fileSystemFactory, credentialCopier);
+  }
 
   @Test
   public void postProcessAfterInitializationWithCorrectBeanName() throws Exception {
```
patduin: I am not a fan of this test because it is not really testing what the code is supposed to do, you're only testing that a `newPath` is called but that's not the whole picture.
Just going on the branches in the code , you should have three tests:
1) a test where the bean name doesn't match, you got that one (`postProcessAfterInitializationWithIncorrectBeanName`), good!
2) a test where bean is good and path is not null (almost got that `postProcessorWithNonNullCredentialProvider`)
3) a test where bean is good and path is null (almost got that `postProcessAfterInitializationWithCorrectBeanName`)

I think they can be a bit more clear and verify better what the code is doing.
Consider what you have now for 2,3 and replace with:
```
  @Test
  public void postProcessAfterInitializationWithConfigurationBeanProviderPathIsNotNull() throws Exception {
    String beanName = CommonBeans.BEAN_BASE_CONF;
    when(credentialPathProvider.newPath()).thenReturn(new Path("/test.json"));
    when(fileSystemFactory.getFileSystem(configuration)).thenReturn(fileSystem);

    processor.postProcessAfterInitialization(configuration, beanName);
    verify(bindGoogleHadoopFileSystem).bindFileSystem(configuration);
    verify(credentialCopier)
        .copyCredentials(fileSystem, configuration, credentialPathProvider, distributedFileSystemPathProvider);
  }

  @Test
  public void postProcessAfterInitializationWithConfigurationBeanProviderPathIsNull() throws Exception {
    String beanName = CommonBeans.BEAN_BASE_CONF;
    when(credentialPathProvider.newPath()).thenReturn(null);

    processor.postProcessAfterInitialization(configuration, beanName);
    verifyZeroInteractions(bindGoogleHadoopFileSystem, credentialCopier, fileSystemFactory);
  }
```


If you prefer `doReturn.when` instead of `when.thenReturn` I am ok with that, you'll see the latter a bit more in our code though (as you get compile time type checking and imho it reads better.) I am sure there is a whole discussion somewhere on the internet favoring one or the other. Either way I can read it.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/context/GCPBeanPostProcessorTest.java
```diff
@@ -16,46 +16,78 @@
...
   }
-}
\ No newline at end of file
+
+  @Test
+  public void postProcessorWithNonNullCredentialProvider() {
+    String beanName = CommonBeans.BEAN_BASE_CONF;
+    doReturn(new Path("/test.json")).when(credentialPathProvider).newPath();
+    processor.postProcessAfterInitialization(configuration, beanName);
+    verify(credentialPathProvider).newPath();
```
patduin: In general: if you have a `when(credentialPathProvider).newPath()` you don't need a `verify(credentialPathProvider).newPath()` as the verify will be superfluous.

circus-train-gcp/src/test/java/com/hotels/bdp/circustrain/gcp/GCPCredentialCopierTest.java
```diff
@@ -15,60 +15,65 @@
...
-    copier.copyCredentials();
-    verify(fs, times(1)).copyFromLocalFile(any(Path.class), any(Path.class));
-    assertNotNull(conf.get("mapreduce.job.cache.files"));
+  public void copyCredentialsWithCredentialProviderSupplied() throws Exception {
+    copier.copyCredentials(fileSystem, conf, credentialPathProvider, distributedFileSystemPathProvider);
+    verify(fileSystem).copyFromLocalFile(credentialsFileRelativePath, dfsAbsolutePath);
+    verify(fileSystem).deleteOnExit(dfsDirectory);
+    assertNotNull(conf.get(DISTRIBUTED_CACHE_PROPERTY));
+    assertThat(conf.get(DISTRIBUTED_CACHE_PROPERTY), is(dfsAbsolutePath + SYMLINK_FLAG + credentialsFileRelativePath));
+    assertFalse(fileSystem.exists(dfsAbsolutePath));
```
patduin: remove, you are asserting a mocked object

PR #89, opened 2018-09-28
Change banner character colour

AnanaMJ:
Banner before:
image

Banner now:
image


massdosage:
You are definitely on the right team to have noticed this and to care enough to change it!

abhimanyugupta07:
Where's the changeLog entry ? 😄

AnanaMJ:
> @abhimanyugupta07 Where's the changeLog entry ? 😄

Should I have one? If so, what would should it say? 😆

PR #96, opened 2018-11-15
Reduce scope of component scanning

### LINKED ISSUE #95: Component scanning of base circustrain package too wide
This is related to https://github.com/HotelsDotCom/circus-train-bigquery/issues/20

The current component scanning in Circus Train uses `com.hotels.bdp.circustrain` as the base package. This means that if other libraries are placed in the classpath with components under this package then Circus Train will attempt to load any configuration/components etc. While this is usually the desired behaviour, sometimes these additional components only work if certain configuration is present. This means that one can't then use the same installation of Circus Train for replications that require these extensions and those that don't. 

One solution is to keep the extension files in a separate location (i.e. not the main Circus Train lib folder) and only add them to the classpath for replications that require them. However it would be good if Circus Train was also changed to be more restrictive in terms of the base packages it scans and thus forces extensions to clearly indicate that they are extensions using the `extensions-package` mechanism.

**Acceptance Criteria**
* Circus Train base component package updated to be the bare minimum required for running Circus Train without any extensions.
* The `circus-train-aws-sns` extension updated to make use of the standard extension mechanism instead of it being implicit via a sub-package.
* All relevant documentation updated.

massdosage:


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ### [TBD] - TBD
 ### Changed
+* Narrowed component scanning to be internal base packages instead of `com.hotels.bdp.circustrain`. See (#95)[https://github.com/HotelsDotCom/circus-train/issues/95]. Note this change is _not_ backwards compatible for any Circus Train extensions that are in the `com.hotels.bdp.circustrain` package - these were in effect being
```
AnanaMJ: The brackets should be the other way around:
`[#95](https://github.com/HotelsDotCom/circus-train/issues/95)` is [#95](https://github.com/HotelsDotCom/circus-train/issues/95)
but
`(#95)[https://github.com/HotelsDotCom/circus-train/issues/95]` is (#95)[https://github.com/HotelsDotCom/circus-train/issues/95]

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ### [TBD] - TBD
 ### Changed
+* Narrowed component scanning to be internal base packages instead of `com.hotels.bdp.circustrain`. See (#95)[https://github.com/HotelsDotCom/circus-train/issues/95]. Note this change is _not_ backwards compatible for any Circus Train extensions that are in the `com.hotels.bdp.circustrain` package - these were in effect being 
+implicitly scanned and loaded but won't be now. Instead these extensions will now need to be added using Circus Train's (standard extension loading mechanism)[https://github.com/HotelsDotCom/circus-train#loading-extensions]. In particular this applies to the "reference" (AWS SNS event listener extension)[https://github.com/HotelsDotCom/circus-train/tree/master/circus-train-aws-sns].
```
AnanaMJ: Same as above

massdosage:
Fixes #95

PR #101, opened 2018-12-11
Moved HK readme and added code to configure batch size

patduin:
Related HK PR: https://github.com/HotelsDotCom/housekeeping/pull/42

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,12 +456,10 @@ Housekeeping is the process that removes expired and orphaned data on the replic
 
 |Property|Required|Description|
 |----|----|----|
-|`housekeeping.h2.database`|No|The location of the h2 housekeeping database. Defaults to `${instance.home}/data/${instance.name}/housekeeping`.|
-|`housekeeping.expired-path-duration`|No|Time To Live (TTL) of legacy replica paths in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) format: only days, hours, minutes and seconds can be specified in the expression.|
-|`housekeeping.data-source.driver-class-name`|No|Java classname of the database JDBC driver.|
-|`housekeeping.data-source.url`|No|JDBC connection URL. Refer to the [Instance Globals](#instance-globals) section for more details.|
-|`housekeeping.data-source.username`|No|database user with access to `circus_train` schema.|
-|`housekeeping.data-source.password`|No|database user's password.|
+|`housekeeping.h2.database`|No|The location of the h2 housekeeping database. Defaults to `${instance.home}/data/${instance.name}/housekeeping`. Refer to the [Instance Globals](#instance-globals) section for more details.|
+|`housekeeping.schema-name`|No|Database schema name to use. Circus Train default: 'circus_train'|
+
+For more details on Housekeeping configuration, including example on how to override the H2 default Database with something a bit more robust, please consult the [housekeeping documentation](https://github.com/HotelsDotCom/housekeeping).
 
 
 ## Metric Reporting
--- a/README.md
+++ b/README.md
@@ -459,7 +459,7 @@ Housekeeping is the process that removes expired and orphaned data on the replic
 |`housekeeping.h2.database`|No|The location of the h2 housekeeping database. Defaults to `${instance.home}/data/${instance.name}/housekeeping`. Refer to the [Instance Globals](#instance-globals) section for more details.|
 |`housekeeping.schema-name`|No|Database schema name to use. Circus Train default: 'circus_train'|
 
-For more details on Housekeeping configuration, including example on how to override the H2 default Database with something a bit more robust, please consult the [housekeeping documentation](https://github.com/HotelsDotCom/housekeeping).
+For more details on Housekeeping configuration, including examples on how to override the H2 default database with something a bit more robust, please consult the [Housekeeping documentation](https://github.com/HotelsDotCom/housekeeping).
 
 
 ## Metric Reporting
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -456,12 +456,10 @@ Housekeeping is the process that removes expired and orphaned data on the replic
...
 |Property|Required|Description|
 |----|----|----|
-|`housekeeping.h2.database`|No|The location of the h2 housekeeping database. Defaults to `${instance.home}/data/${instance.name}/housekeeping`.|
-|`housekeeping.expired-path-duration`|No|Time To Live (TTL) of legacy replica paths in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) format: only days, hours, minutes and seconds can be specified in the expression.|
-|`housekeeping.data-source.driver-class-name`|No|Java classname of the database JDBC driver.|
-|`housekeeping.data-source.url`|No|JDBC connection URL. Refer to the [Instance Globals](#instance-globals) section for more details.|
-|`housekeeping.data-source.username`|No|database user with access to `circus_train` schema.|
-|`housekeeping.data-source.password`|No|database user's password.|
+|`housekeeping.h2.database`|No|The location of the h2 housekeeping database. Defaults to `${instance.home}/data/${instance.name}/housekeeping`. Refer to the [Instance Globals](#instance-globals) section for more details.|
+|`housekeeping.schema-name`|No|Database schema name to use. Circus Train default: 'circus_train'|
```
massdosage: So we're keeping this one here because we override the default value for circus_train right?

README.md
```diff
@@ -456,12 +456,10 @@ Housekeeping is the process that removes expired and orphaned data on the replic
 
 |Property|Required|Description|
 |----|----|----|
-|`housekeeping.h2.database`|No|The location of the h2 housekeeping database. Defaults to `${instance.home}/data/${instance.name}/housekeeping`.|
-|`housekeeping.expired-path-duration`|No|Time To Live (TTL) of legacy replica paths in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) format: only days, hours, minutes and seconds can be specified in the expression.|
-|`housekeeping.data-source.driver-class-name`|No|Java classname of the database JDBC driver.|
-|`housekeeping.data-source.url`|No|JDBC connection URL. Refer to the [Instance Globals](#instance-globals) section for more details.|
-|`housekeeping.data-source.username`|No|database user with access to `circus_train` schema.|
-|`housekeeping.data-source.password`|No|database user's password.|
+|`housekeeping.h2.database`|No|The location of the h2 housekeeping database. Defaults to `${instance.home}/data/${instance.name}/housekeeping`. Refer to the [Instance Globals](#instance-globals) section for more details.|
```
massdosage: This is Circus Train specific and isn't actually in Housekeeping?

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/CircusTrainHousekeepingConfiguration.java
```diff
@@ -39,8 +40,10 @@
   @Bean
   HousekeepingService housekeepingService(
       LegacyReplicaPathRepository legacyReplicaPathRepository,
-      @Qualifier("baseConf") org.apache.hadoop.conf.Configuration baseConf) {
-    return new FileSystemHousekeepingService(legacyReplicaPathRepository, baseConf);
+      @Qualifier("baseConf") org.apache.hadoop.conf.Configuration baseConf,
+      Housekeeping housekeeping) {
+    return new FileSystemHousekeepingService(legacyReplicaPathRepository, baseConf,
```
massdosage: Doesn't this require a bump in the version of HK that CT depends on?

README.md
```diff
@@ -456,12 +456,10 @@ Housekeeping is the process that removes expired and orphaned data on the replic
...
+|`housekeeping.schema-name`|No|Database schema name to use. Circus Train default: 'circus_train'|
+
+For more details on Housekeeping configuration, including example on how to override the H2 default Database with something a bit more robust, please consult the [housekeeping documentation](https://github.com/HotelsDotCom/housekeeping).
```
massdosage: example  -> examples

massdosage: housekeeping documentation -> Housekeeping documentation

massdosage: Database -> database

massdosage: https://github.com/HotelsDotCom/housekeeping <- can you link directly to the configuration section in the README? I think that will be a bit clearer?

README.md
patduin: oh it's a bit complex:
`housekeeping.h2.database` is in housekeeping But the reference to the instance globals isn't. I don't really like that in housekeeping tbh, but in housekeeping we set this as well. Perhaps I just copy that, I'll sleep on this

README.md
patduin: yes

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/CircusTrainHousekeepingConfiguration.java
patduin: yes the PR is on a branch which already bumps it, I'll merge the other one in first as soon as HK is released

README.md
patduin: I prefer to just point to the documentation, you know we'll edit it at some point and break it + the whole readme is more or less about sorting out the connection.

README.md
patduin: I've added an entry on this in the HK readme as well, keeping this here as it makes more sense in this context imo

circus-train-housekeeping/src/main/java/com/hotels/bdp/circustrain/housekeeping/CircusTrainHousekeepingConfiguration.java
massdosage: OK, makes sense.

README.md
massdosage: OK, so we keep the ones most often used by CT duplicated here and then provide a link to the rest for those who might want to customise it further? I'm OK with this, let's just try get the text to be the same (i.e. copy over from HK PR once that ones merged).

README.md
massdosage: OK

README.md
patduin: No the text shouldn't be the same because in CT context `instance.home` for instance has a default value (`user.home`). I think it makes more sense leaving like this reading it as a CT user???

massdosage: OK, yeah, that's fine.

PR #106, opened 2019-01-07
Checkstyle and findbugs refactoring

AnanaMJ:
Checkstyle warnings: 164 -> 45
Findbugs warnings: 44 -> 13

### DIFF
--- a/circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/util/SchemaCopier.java
+++ b/circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/util/SchemaCopier.java
@@ -76,7 +76,7 @@ public class SchemaCopier {
       temporaryDirectory = Files.createTempDirectory("avro-schema-download-folder");
       temporaryDirectory.toFile().deleteOnExit();
     } catch (IOException e) {
-      throw new CircusTrainException("Couldn't create temporaryDirectory " + temporaryDirectory, e);
+      throw new CircusTrainException("Couldn't create temporaryDirectory in the default temporary directory", e);
     }
     return temporaryDirectory;
   }
--- a/circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AWSConstants.java
+++ b/circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AWSConstants.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Expedia Inc.
+ * Copyright (C) 2016-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,7 +23,7 @@ public final class AWSConstants {
   public static final String FS_PROTOCOL_S3 = "s3";
   public static final String FS_PROTOCOL_S3N = "s3n";
   public static final String FS_PROTOCOL_S3A = "s3a";
-  public static final String[] S3_FS_PROTOCOLS = { FS_PROTOCOL_S3, FS_PROTOCOL_S3N, FS_PROTOCOL_S3A };
+  static final String[] S3_FS_PROTOCOLS = { FS_PROTOCOL_S3, FS_PROTOCOL_S3N, FS_PROTOCOL_S3A };
 
   private AWSConstants() {}
 }
--- a/circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/wrappers/PathMetadata.java
+++ b/circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/wrappers/PathMetadata.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Expedia Inc.
+ * Copyright (C) 2016-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -50,7 +50,7 @@ public class PathMetadata implements Serializable {
       checkSumLength = checksum.getLength();
       this.checksum = checksum.getBytes();
     }
-    this.childrenMetadata = childrenMetadata == null ? ImmutableList.<PathMetadata> of()
+    this.childrenMetadata = childrenMetadata == null ? ImmutableList.<PathMetadata>of()
         : ImmutableList.copyOf(childrenMetadata);
   }
 
--- a/circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/wrappers/PathMetadata.java
+++ b/circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/wrappers/PathMetadata.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2018 Expedia Inc.
+ * Copyright (C) 2016-2019 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 package com.hotels.bdp.circustrain.comparator.hive.wrappers;
 
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.List;
 
 import org.apache.hadoop.fs.FileChecksum;
@@ -71,7 +72,11 @@ public class PathMetadata implements Serializable {
   }
 
   public byte[] getChecksum() {
-    return checksum;
+    if (checksum == null) {
+      return new byte[] {};
+    } else {
+      return Arrays.copyOf(checksum, checksum.length);
+    }
   }
 
   public List<PathMetadata> getChildrenMetadata() {
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2018 Expedia Inc.
+ * Copyright (C) 2016-2019 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
@@ -115,8 +115,9 @@ public class CircusTrain {
       printCircusTrainHelp(e.getErrors());
     } catch (BeanCreationException e) {
       LOG.error(e.getMessage(), e);
-      if (e.getMostSpecificCause() instanceof BindException) {
-        printCircusTrainHelp(((BindException) e.getMostSpecificCause()).getAllErrors());
+      Throwable mostSpecificCause = e.getMostSpecificCause();
+      if (mostSpecificCause instanceof BindException) {
+        printCircusTrainHelp(((BindException) mostSpecificCause).getAllErrors());
       }
     }
 
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
@@ -86,8 +86,7 @@ import com.hotels.bdp.circustrain.manifest.ManifestAttributes;
     "com.hotels.bdp.circustrain.housekeeping",
     "com.hotels.bdp.circustrain.metrics.conf",
     "com.hotels.bdp.circustrain.s3mapreducecpcopier",
-    "com.hotels.bdp.circustrain.s3s3copier"
-})
+    "com.hotels.bdp.circustrain.s3s3copier" })
 public class CircusTrain {
   private static final Logger LOG = LoggerFactory.getLogger(CircusTrain.class);
 
@@ -98,18 +97,19 @@ public class CircusTrain {
     int exitCode = -1;
     try {
       String defaultModules = Joiner.on(",").join(Modules.REPLICATION, Modules.HOUSEKEEPING);
-      exitCode = SpringApplication.exit(new SpringApplicationBuilder(CircusTrain.class)
-          .properties("spring.config.location:${config:null}")
-          .properties("spring.profiles.active:${modules:" + defaultModules + "}")
-          .properties("instance.home:${user.home}")
-          .properties("instance.name:${source-catalog.name}_${replica-catalog.name}")
-          .properties("jasypt.encryptor.password:${password:null}")
-          .properties("housekeeping.schema-name:circus_train")
-          .registerShutdownHook(true)
-          .initializers(new ExtensionInitializer())
-          .listeners(new ConfigFileValidationApplicationListener())
-          .build()
-          .run(args));
+      exitCode = SpringApplication
+          .exit(new SpringApplicationBuilder(CircusTrain.class)
+              .properties("spring.config.location:${config:null}")
+              .properties("spring.profiles.active:${modules:" + defaultModules + "}")
+              .properties("instance.home:${user.home}")
+              .properties("instance.name:${source-catalog.name}_${replica-catalog.name}")
+              .properties("jasypt.encryptor.password:${password:null}")
+              .properties("housekeeping.schema-name:circus_train")
+              .registerShutdownHook(true)
+              .initializers(new ExtensionInitializer())
+              .listeners(new ConfigFileValidationApplicationListener())
+              .build()
+              .run(args));
     } catch (ConfigFileValidationException e) {
       LOG.error(e.getMessage(), e);
       printCircusTrainHelp(e.getErrors());
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/PartitionedTableMetadataMirrorReplication.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/PartitionedTableMetadataMirrorReplication.java
@@ -1,9 +1,5 @@
 /**
-<<<<<<< HEAD
- * Copyright (C) 2016-2018 Expedia Inc.
-=======
  * Copyright (C) 2016-2019 Expedia Inc.
->>>>>>> master
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/PartitionedTableMetadataMirrorReplication.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/PartitionedTableMetadataMirrorReplication.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Expedia Inc.
+ * Copyright (C) 2016-2018 Expedia Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -79,7 +79,7 @@ class PartitionedTableMetadataMirrorReplication implements Replication {
 
       // We expect all partitions to be under the table base path
       SourceLocationManager sourceLocationManager = source.getLocationManager(sourceTable, sourcePartitions, eventId,
-          Collections.<String, Object> emptyMap());
+          Collections.<String, Object>emptyMap());
       ReplicaLocationManager replicaLocationManager = new MetadataMirrorReplicaLocationManager(sourceLocationManager,
           TableType.PARTITIONED);
       sourceLocationManager.cleanUpLocations();
@@ -88,8 +88,8 @@ class PartitionedTableMetadataMirrorReplication implements Replication {
         replica.updateMetadata(eventId, sourceTableAndStatistics, replicaDatabaseName, replicaTableName,
             replicaLocationManager);
         LOG.info(
-            "No matching partitions found on table {}.{} with predicate {}. Table metadata updated, no partitions were updated.",
-            database, table, partitionPredicate);
+            "No matching partitions found on table {}.{} with predicate {}. "+
+            "Table metadata updated, no partitions were updated.", database, table, partitionPredicate);
       } else {
         replica.updateMetadata(eventId, sourceTableAndStatistics, sourcePartitionsAndStatistics, sourceLocationManager,
             replicaDatabaseName, replicaTableName, replicaLocationManager);
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
@@ -86,10 +86,10 @@ public class ReplicationFactoryImpl implements ReplicationFactory {
     Replication replication = null;
     if (partitionKeys == null || partitionKeys.isEmpty()) {
       replication = createUnpartitionedTableReplication(tableReplication, source, replica, sourceDatabaseName,
-          sourceTableName, replicaDatabaseName, replicaTableName, replicaTableLocation, replication);
+          sourceTableName, replicaDatabaseName, replicaTableName, replicaTableLocation);
     } else {
       replication = createPartitionedTableReplication(tableReplication, source, replica, sourceDatabaseName,
-          sourceTableName, replicaDatabaseName, replicaTableName, replicaTableLocation, replication);
+          sourceTableName, replicaDatabaseName, replicaTableName, replicaTableLocation);
     }
     return replication;
   }
@@ -102,8 +102,8 @@ public class ReplicationFactoryImpl implements ReplicationFactory {
       String sourceTableName,
       String replicaDatabaseName,
       String replicaTableName,
-      String replicaTableLocation,
-      Replication replication) {
+      String replicaTableLocation) {
+    Replication replication = null;
     PartitionPredicate partitionPredicate = partitionPredicateFactory.newInstance(tableReplication);
     switch (tableReplication.getReplicationMode()) {
     case METADATA_MIRROR:
@@ -136,8 +136,8 @@ public class ReplicationFactoryImpl implements ReplicationFactory {
       String sourceTableName,
       String replicaDatabaseName,
       String replicaTableName,
-      String replicaTableLocation,
-      Replication replication) {
+      String replicaTableLocation) {
+    Replication replication = null;
     switch (tableReplication.getReplicationMode()) {
     case METADATA_MIRROR:
       replication = new UnpartitionedTableMetadataMirrorReplication(sourceDatabaseName, sourceTableName, source,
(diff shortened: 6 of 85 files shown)


circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/wrappers/PathMetadata.java
```diff
@@ -71,7 +72,11 @@ public int getChecksumLength() {
   }
 
   public byte[] getChecksum() {
-    return checksum;
+    if (checksum == null) {
+      return new byte[] {};
+    } else {
+      return Arrays.copyOf(checksum, checksum.length);
+    }
```
AnanaMJ: This was done because of:
```
com.hotels.bdp.circustrain.comparator.hive.wrappers.PathMetadata.getChecksum() may expose internal 
representation by returning PathMetadata.checksum

Returning a reference to a mutable object value stored in one of the object's fields exposes the internal 
representation of the object. If instances are accessed by untrusted code, and unchecked changes to the mutable 
object would compromise security or other important properties, you will need to do something different. Returning 
a new copy of the object is better approach in many situations.
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/CircusTrain.java
```diff
@@ -98,25 +97,27 @@ public static void main(String[] args) throws Exception {
...
     } catch (ConfigFileValidationException e) {
       LOG.error(e.getMessage(), e);
       printCircusTrainHelp(e.getErrors());
     } catch (BeanCreationException e) {
       LOG.error(e.getMessage(), e);
-      if (e.getMostSpecificCause() instanceof BindException) {
-        printCircusTrainHelp(((BindException) e.getMostSpecificCause()).getAllErrors());
+      Throwable mostSpecificCause = e.getMostSpecificCause();
+      if (mostSpecificCause instanceof BindException) {
+        printCircusTrainHelp(((BindException) mostSpecificCause).getAllErrors());
```
AnanaMJ: Because we get warnings about unconfirmed casts (Findbugs can't figure out that we are checking if they are of the right instance), I have refactored it this way.
I think getting warnings about unconfirmed casts is important, so I preferred to change the code rather than silence the warning.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
```diff
@@ -86,10 +86,10 @@ public Replication newInstance(TableReplication tableReplication) {
     Replication replication = null;
     if (partitionKeys == null || partitionKeys.isEmpty()) {
       replication = createUnpartitionedTableReplication(tableReplication, source, replica, sourceDatabaseName,
-          sourceTableName, replicaDatabaseName, replicaTableName, replicaTableLocation, replication);
+          sourceTableName, replicaDatabaseName, replicaTableName, replicaTableLocation);
```
AnanaMJ: `replication` is initialised with `null` and then gets the value from `createUnpartitionedTableReplication` or `createPartitionedTableReplication`, which means that the value does not need to change by reference.

circus-train-hive-view/src/main/java/com/hotels/bdp/circustrain/hive/view/transformation/TableProcessor.java
```diff
@@ -35,12 +35,15 @@
...
+      astNode = (ASTNode) node;
+    } else {
+      throw new ClassCastException("Cannot cast Node to ASTNode");
+    }
     if (astNode.getToken() != null && astNode.getToken().getText() != null) {
-      switch (astNode.getToken().getText()) {
-      case "TOK_TABNAME":
+      if ("TOK_TABNAME".equals(astNode.getToken().getText())) {
         tables.add(extractTableName(astNode));
-        break;
```
AnanaMJ: This was a switch statement with one case and no `default`, so it was replaced by an if statement.

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/S3MapReduceCpOptions.java
```diff
@@ -285,11 +301,11 @@ void setS3ServerSideEncryption(boolean s3ServerSideEncryption) {
   }
 
   public String getStorageClass() {
-    return storageClass.toString();
+    return storageClass;
```
AnanaMJ: `storageClass` is already a String.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AWSConstants.java
```diff
@@ -23,7 +23,7 @@
   public static final String FS_PROTOCOL_S3 = "s3";
   public static final String FS_PROTOCOL_S3N = "s3n";
   public static final String FS_PROTOCOL_S3A = "s3a";
-  public static final String[] S3_FS_PROTOCOLS = { FS_PROTOCOL_S3, FS_PROTOCOL_S3N, FS_PROTOCOL_S3A };
+  static final String[] S3_FS_PROTOCOLS = { FS_PROTOCOL_S3, FS_PROTOCOL_S3N, FS_PROTOCOL_S3A };
```
patduin: This is not used CT big query or some other external module?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
```diff
@@ -51,33 +51,67 @@
   private ReplicationState replicationState = new ReplicationState();
 
   @VisibleForTesting
-  class ReplicationState {
-    List<String> partitionKeys;
-    int partitionsAltered;
-    long bytesReplicated;
+  static class ReplicationState {
```
patduin: I think I prefer left side. I this again a complaint that the inner class doesn't have a reference to the outer class? There were loads of references to the inner class though.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/MetricsListener.java
```diff
@@ -80,8 +80,10 @@ public void tableReplicationFailure(EventTableReplication tableReplication, Stri
 
   @Override
   public void copierEnd(Metrics metrics) {
-    runningMetricsReporter.report();
-    runningMetricsReporter.stop();
+    if (runningMetricsReporter != null) {
```
patduin: This hides a bug maybe better throw an IllegalStateException when null?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/util/LibJarDeployer.java
```diff
@@ -42,15 +43,32 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(LibJarDeployer.class);
 
+  private static class PriviledgedClassLoader implements PrivilegedAction<Void> {
```
patduin: What the reason for this one? The interface javadoc seems to not really justify the usage here

circus-train-hive-view/src/main/java/com/hotels/bdp/circustrain/hive/view/transformation/TableProcessor.java
```diff
@@ -35,12 +35,15 @@
   @Override
   public Object process(Node node, Stack<Node> stack, NodeProcessorCtx procCtx, Object... nodeOutputs)
     throws SemanticException {
-    ASTNode astNode = (ASTNode) node;
+    ASTNode astNode;
```
patduin: I prefer left, this is really satisfying checkstyle for the sake of it I think

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/mapreduce/RetriableFileCopyCommand.java
```diff
@@ -133,10 +134,9 @@ private long doCopy(Mapper.Context context, FileStatus sourceFileStatus, S3Uploa
   }
 
   private static ThrottledInputStream getInputStream(Path path, Configuration conf) throws IOException {
-    try {
-      FileSystem fs = path.getFileSystem(conf);
-      long bandwidthMB = conf.getInt(ConfigurationVariable.MAX_BANDWIDTH.getName(),
-          ConfigurationVariable.MAX_BANDWIDTH.defaultIntValue());
+    try (FileSystem fs = path.getFileSystem(conf)) {
```
patduin: We use `FileSystem` all over the place and we never call close I think, is this the only place where checkstyle is complaining about this, is that because of the `fs.open` call on line 140? Getting the FS from the path might involve getting a cached one and closing it would remove it from the cache I don't think closing it is necessary in our case.

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/mapreduce/RetriableFileCopyCommand.java
```diff
@@ -159,8 +159,9 @@ private Transfer startTransfer(Mapper.Context context, S3UploadDescriptor upload
           input, uploadDescriptor.getMetadata());
       // We add 1 to the buffer size as per the com.amazonaws.RequestClientOptions doc
       request.getRequestClientOptions().setReadLimit(bufferSize + 1);
+      input.close();
```
patduin: try with resource on line 148 maybe better?

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/mapreduce/UniformSizeInputFormat.java
```diff
@@ -125,7 +126,7 @@
   private static Path getListingFilePath(Configuration configuration) {
     final String listingFilePathString = configuration.get(S3MapReduceCpConstants.CONF_LABEL_LISTING_FILE_PATH, "");
 
-    assert !listingFilePathString.equals("") : "Couldn't find listing file. Invalid input.";
+    assert !"".equals(listingFilePathString) : "Couldn't find listing file. Invalid input.";
```
patduin: Remove the assert or throw an illegalArgumentException if this is really bad. I'm fine with removing the line, this is code I think we copied from hadoop so probably ok to keep as is.

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/mapreduce/lib/DynamicInputChunk.java
```diff
@@ -68,7 +69,7 @@ private static void initializeChunkInvariants(Configuration config) throws IOExc
 
   private static String getListingFilePath(Configuration configuration) {
     final String listingFileString = configuration.get(S3MapReduceCpConstants.CONF_LABEL_LISTING_FILE_PATH, "");
-    assert !listingFileString.equals("") : "Listing file not found.";
+    assert !"".equals(listingFileString) : "Listing file not found.";
```
patduin: same as above

circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/wrappers/PathMetadata.java
```diff
@@ -71,7 +72,11 @@ public int getChecksumLength() {
   }
 
   public byte[] getChecksum() {
-    return checksum;
+    if (checksum == null) {
+      return new byte[] {};
```
massdosage: The old behaviour would have returned `null` here so we need to be careful that all calling code can deal with this change (or return null again).

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
```diff
@@ -93,22 +127,24 @@ public void circusTrainStartUp(String[] args, EventSourceCatalog sourceCatalog,
...
   @Override
   public void resolvedMetaStoreSourceTable(EventTable table) {
-    replicationState.partitionKeys = table.getPartitionKeys();
+    replicationState.setPartitionKeys(table.getPartitionKeys());
   }
 
   @Override
   public void partitionsToCreate(EventPartitions partitions) {
-    replicationState.partitionsAltered += partitions.getEventPartitions().size();
+    replicationState
```
massdosage: If we do decide to stick with this approach then having an `incrementPartitionsAltered` method which takes the new size and then adds it internally to the new size would probably be easier to use.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/MetricsListener.java
massdosage: +1

circus-train-distcp-copier/src/main/java/com/hotels/bdp/circustrain/distcpcopier/DistCpCopier.java
```diff
@@ -56,6 +56,27 @@ public Job exec(Configuration conf, DistCpOptions options) throws Exception {
...
+
+    public JobCounter(Job job, String counter) {
+      this.job = job;
+      this.counter = counter;
+    }
+
+    @Override
+    public Long getValue() {
+      try {
+        return job.getCounters().findCounter(FileSystemCounter.class.getName(), counter).getValue();
```
massdosage: This will only find Counters with a FileSystemCounter group name so a more accurate name for the class would be `FileSystemCounterGauge`

circus-train-hive-view/src/main/java/com/hotels/bdp/circustrain/hive/view/transformation/TableProcessor.java
massdosage: I agree, especially since we're throwing the exception that Java would throw anyway. I wonder if we should turn this one off? Generally if you're forcing a cast then you at least think you know what you're doing. What's the exact warning that checkstyle gives here?

circus-train-hive-view/src/main/java/com/hotels/bdp/circustrain/hive/view/transformation/TableProcessor.java
massdosage: OK, maybe the intention was to support more later, what you've done is fine, it can always be converted back later if more token types are needed.

circus-train-s3-mapreduce-cp-copier/src/main/java/com/hotels/bdp/circustrain/s3mapreducecpcopier/S3MapReduceCpCopier.java
```diff
@@ -57,6 +57,26 @@ public Job exec(Configuration conf, S3MapReduceCpOptions options) throws Excepti
 
   private static final Logger LOG = LoggerFactory.getLogger(S3MapReduceCpCopier.class);
 
+  private static class JobCounter implements Gauge<Long> {
```
massdosage: I think this could be factored into a more generic `JobCounterGauge` which has an additional constructor which takes the group name. If this is set it uses it in the call to `findCounter`, if not it does the below. Then you can replace the above `FileSystemCounterGauge` with it too.

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/mapreduce/UniformSizeInputFormat.java
massdosage: I agree with removing the `assert`, I guess it all depends on what happens if you do new Path("") and if that's something we really want to avoid (if so then IAE is a good alternative).

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -55,6 +55,49 @@
...
+        // so just printing total bytes when completed.
+        LOG
+            .debug("copied object from '{}/{}' to '{}/{}': {} bytes transferred", s3ObjectSummary.getBucketName(),
+                s3ObjectSummary.getKey(), targetS3Uri.getBucket(), targetKey,
+                transfer.getProgress().getTotalBytesToTransfer());
+      }
+    }
+  }
+
+  private static class BytesGauge implements Gauge<Long> {
```
massdosage: This should be an `AtomicLong` gauge as it's got nothing to do with bytes.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -55,6 +55,49 @@
...
+
+    private final AtomicLong bytesReplicated;
```
massdosage: bytesReplicated -> value (then this becomes more clearly generic)

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/util/SchemaCopier.java
```diff
@@ -76,7 +76,7 @@ public Path copy(String source, String destination) {
       temporaryDirectory = Files.createTempDirectory("avro-schema-download-folder");
       temporaryDirectory.toFile().deleteOnExit();
     } catch (IOException e) {
-      throw new CircusTrainException("Couldn't create temporaryDirectory " + temporaryDirectory, e);
+      throw new CircusTrainException("Couldn't create temporaryDirectory in the default temporary directory", e);
```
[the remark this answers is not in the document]
AnanaMJ: So the only place where IOException is thrown is on the line where `createTempDirectory()` is called, and before that, `temporaryDirectory` is null. Wouldn't `temporaryDirectory` still be null if an IOException occurs?

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AWSConstants.java
AnanaMJ: Is searching on GitHub for "S3_FS_PROTOCOLS" enough? I searched for it in circus-train-bigquery and housekeeping and there were no matches for that.

circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/wrappers/PathMetadata.java
AnanaMJ: You make a good point. I'll change it to return `null`.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
AnanaMJ: No, this seems to be related to enforcing encapsulation and it comes from Checkstyle: http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/design/VisibilityModifierCheck.html

AnanaMJ: But I do think that a shorter version is more readable, however there's a warning for each variable that doesn't have a setter and a getter.

patduin: hmm from that link it should complain because of the annotation?

> ignoreAnnotationCanonicalNames - the list of annotations canonical names which ignore variables in consideration, if user will provide short annotation name that type will match to any named the same type without consideration of package, list by default:
> 
> org.junit.Rule
> org.junit.ClassRule
> com.google.common.annotations.VisibleForTesting

the whole point was to not have a getter/setter to make the code a bit more concise :)

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AWSConstants.java
patduin: yeah that should be ok, thanks!

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/util/LibJarDeployer.java
AnanaMJ: Ah, yes, I should've added a comment for this:

```
com.hotels.bdp.circustrain.core.util.LibJarDeployer.libjars(Configuration, Class[]) 
creates a java.net.URLClassLoader classloader, which should be performed within 
a doPrivileged block.

This code creates a classloader, which needs permission if a security manage is 
installed. If this code might be invoked by code that does not have security 
permissions, then the classloader creation needs to occur inside a doPrivileged 
block.
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
AnanaMJ: Ah, yes, then maybe checkstyle could be updated to not warn us about things with `@VisibleForTesting`?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
AnanaMJ: I think Patrick makes a good point about it being more readable before. But it sounds like a good idea, so should I still add in with an `incrementPartitionsAltered` method?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
massdosage: Well, I think generally we would want it to check methods annotated with `@VisibleForTesting` as often these are just "normal" methods/classes that we want to test directly. @patduin I don't think it's complaining because of the annotation it's because the class is visible outside of this class and it exposes its state so a caller from outside can fiddle with its internals which is frowned upon. I'd be OK with extracting this out into its own class and then leaving with the getters/setters.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
massdosage: Sure.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/util/LibJarDeployer.java
massdosage: Interesting, sounds like a reasonable change, we just need to make sure this still works for real when run on a cluster etc.

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/mapreduce/UniformSizeInputFormat.java
AnanaMJ: Im refactoring this to 
```
if ("".equals(listingFilePathString)) {
  throw new IllegalArgumentException("Couldn't find listing file. Invalid input.");
}
```
And I will do similarly for the next ones.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/PartitionedTableMetadataMirrorReplication.java
```diff
@@ -1,5 +1,9 @@
 /**
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
AnanaMJ: Thanks! I didn't see it in the IDE because it was collapsed.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
patduin: They way I read it was it should not complain because of the annotation.
When would use the annotation on a private member? But reading the  link again it says variables so perhaps methods don't apply. Dunno. if it is private I guess checkstyle won't complain but then if becomes harder to test.
How about we make the ReplicationState class private and expose the fields in the listener like so:
```
  List<String> getPartitionKeys() {
    return Collections.unmodifiableList(replicationState.partitionKeys);
  }
```
Then we can test the state without exposing the class and we don't need a million lines extra.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/util/LibJarDeployer.java
patduin: I don't think adding an interface like we have now makes it run in a privileged block I think you then need to call it like this???
```
 AccessController.doPrivileged(new PriviledgedClassLoader(conf, libjars));
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/event/LoggingListener.java
AnanaMJ: Having a method that exposes the fields sounds good, but it seems that the parameters are the same in the test, before and after
`listener.tableReplicationStart(tableReplication, "event-id");`

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/util/LibJarDeployer.java
AnanaMJ: Yes, you're right! I missed that part of it.

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/mapreduce/RetriableFileCopyCommand.java
AnanaMJ: Cool, I'll undo this then :)

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/mapreduce/RetriableFileCopyCommand.java
AnanaMJ: It seems to not work, because on line 156, `input` is reassigned another value...

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -55,6 +55,49 @@
 
   private final static Logger LOG = LoggerFactory.getLogger(S3S3Copier.class);
 
+  private static class BytesTransferStateChangeListener implements TransferStateChangeListener {
+
+    private final S3ObjectSummary s3ObjectSummary;
+    private final AmazonS3URI targetS3Uri;
+    private final String targetKey;
+
+    public BytesTransferStateChangeListener(
```
massdosage: Does this need to be public? Could we make it private or default?

AnanaMJ: I made it private

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/S3MapReduceCpOptions.java
```diff
@@ -442,28 +443,53 @@ public void setCannedAcl(String cannedAcl) {
...
-            ", logPath=" + logPath +
-            ", region='" + region + '\'' +
-            ", ignoreFailures=" + ignoreFailures +
-            ", s3EndpointUri=" + s3EndpointUri +
-            ", uploadRetryCount=" + uploadRetryCount +
-            ", uploadRetryDelayMs=" + uploadRetryDelayMs +
-            ", uploadBufferSize=" + uploadBufferSize +
-            ", cannedAcl='" + cannedAcl + '\'' +
-            '}';
+    return "S3MapReduceCpOptions{"
```
patduin: eh? this is horrible

AnanaMJ: It is! It seems that eclipse auto-formats it like this, but I can undo this

PR #114, opened 2019-01-29
Reset partition parameters for SnsListener

### LINKED ISSUE #104: SNS listener doesn't clear partitioned state correctly
When running two replications: one partitioned and one unpartitioned. 
See the: `com.hotels.bdp.circustrain.aws.sns.event.SnsListener`
The partitionKeys and modifiedPartitions are not cleared in between replications and will be incorrectly reported in sns messages for both replications.

AnanaMJ:
Fixes #104

### DIFF
--- a/circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
+++ b/circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
@@ -136,7 +136,6 @@ public class SnsListener implements LocomotiveListener, SourceCatalogListener, R
   public void tableReplicationFailure(EventTableReplication tableReplication, String eventId, Throwable t) {
     try {
       if (startTime == null) {
-
         startTime = clock.getTime();
       }
       String endTime = clock.getTime();
--- a/circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
+++ b/circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
@@ -128,7 +128,7 @@ public class SnsListener implements LocomotiveListener, SourceCatalogListener, R
           getModifiedPartitions(partitionsToAlter, partitionsToCreate), getBytesReplicated(), null);
       publish(config.getSuccessTopic(), message);
     } finally {
-      resetPartitionState();
+      resetState();
     }
   }
 
@@ -147,7 +147,7 @@ public class SnsListener implements LocomotiveListener, SourceCatalogListener, R
           getModifiedPartitions(partitionsToAlter, partitionsToCreate), getBytesReplicated(), t.getMessage());
       publish(config.getFailTopic(), message);
     } finally {
-      resetPartitionState();
+      resetState();
     }
   }
 
@@ -158,10 +158,11 @@ public class SnsListener implements LocomotiveListener, SourceCatalogListener, R
     return 0L;
   }
 
-  private void resetPartitionState() {
+  private void resetState() {
     partitionsToCreate = null;
     partitionsToAlter = null;
     partitionKeyTypes = null;
+    startTime = null;
   }
 
   @Override
--- a/circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
+++ b/circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
@@ -128,10 +128,7 @@ public class SnsListener implements LocomotiveListener, SourceCatalogListener, R
           getModifiedPartitions(partitionsToAlter, partitionsToCreate), getBytesReplicated(), null);
       publish(config.getSuccessTopic(), message);
     } finally {
-      // reset registered partitions between database cycles
-      partitionsToCreate = null;
-      partitionsToAlter = null;
-      partitionKeyTypes = null;
+      resetPartitionState();
     }
   }
 
@@ -151,10 +148,7 @@ public class SnsListener implements LocomotiveListener, SourceCatalogListener, R
           getModifiedPartitions(partitionsToAlter, partitionsToCreate), getBytesReplicated(), t.getMessage());
       publish(config.getFailTopic(), message);
     } finally {
-      // reset registered partitions between database cycles
-      partitionsToCreate = null;
-      partitionsToAlter = null;
-      partitionKeyTypes = null;
+      resetPartitionState();
     }
   }
 
@@ -163,6 +157,12 @@ public class SnsListener implements LocomotiveListener, SourceCatalogListener, R
       return metrics.getBytesReplicated();
     }
     return 0L;
+  }
+
+  private void resetPartitionState() {
+    partitionsToCreate = null;
+    partitionsToAlter = null;
+    partitionKeyTypes = null;
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
```diff
@@ -118,29 +118,44 @@ public void tableReplicationStart(EventTableReplication tableReplication, String
...
+      partitionKeyTypes = null;
+    }
   }
 
   @Override
   public void tableReplicationFailure(EventTableReplication tableReplication, String eventId, Throwable t) {
-    if (startTime == null) {
-      startTime = clock.getTime();
+    try {
+      if (startTime == null) {
```
massdosage: This code was already here but I wonder why we have this null check and change to `startTime` in `tableReplicationFailure` but not in `tableReplicationSuccess`? This makes me think though - should we also set `startTime` to null at the end of both of these methods?

circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
```diff
@@ -118,29 +118,44 @@ public void tableReplicationStart(EventTableReplication tableReplication, String
...
+      EventReplicaTable replicaTable = tableReplication.getReplicaTable();
+      SnsMessage message = new SnsMessage(SnsMessageType.FAILURE, config.getHeaders(), startTime, endTime, eventId,
+          sourceCatalog.getName(), replicaCatalog.getName(), replicaCatalog.getHiveMetastoreUris(),
+          tableReplication.getSourceTable().getQualifiedName(), tableReplication.getQualifiedReplicaName(),
+          replicaTable.getTableLocation(), partitionKeyTypes,
+          getModifiedPartitions(partitionsToAlter, partitionsToCreate), getBytesReplicated(), t.getMessage());
+      publish(config.getFailTopic(), message);
+    } finally {
+      // reset registered partitions between database cycles
+      partitionsToCreate = null;
```
massdosage: I don't think this is covered by any tests?

circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
```diff
@@ -118,29 +118,44 @@ public void tableReplicationStart(EventTableReplication tableReplication, String
...
+      String endTime = clock.getTime();
+      EventReplicaTable replicaTable = tableReplication.getReplicaTable();
+      SnsMessage message = new SnsMessage(SnsMessageType.SUCCESS, config.getHeaders(), startTime, endTime, eventId,
+          sourceCatalog.getName(), replicaCatalog.getName(), replicaCatalog.getHiveMetastoreUris(),
+          tableReplication.getSourceTable().getQualifiedName(), tableReplication.getQualifiedReplicaName(),
+          replicaTable.getTableLocation(), partitionKeyTypes,
+          getModifiedPartitions(partitionsToAlter, partitionsToCreate), getBytesReplicated(), null);
+      publish(config.getSuccessTopic(), message);
+    } finally {
+      // reset registered partitions between database cycles
```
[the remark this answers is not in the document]
AnanaMJ: Yep, all good points :) 
By having a function I won't need the comment so that's even better!

circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
```diff
@@ -118,29 +118,44 @@ public void tableReplicationStart(EventTableReplication tableReplication, String
...
+    }
   }
 
   @Override
   public void tableReplicationFailure(EventTableReplication tableReplication, String eventId, Throwable t) {
-    if (startTime == null) {
-      startTime = clock.getTime();
+    try {
+      if (startTime == null) {
+
```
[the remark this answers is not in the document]
AnanaMJ: Ah, when I added the `try`, Eclipse put the closing bracket on that line 😆
I removed it now

circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
AnanaMJ: Maybe because of the functions are used in `Locomotive.java`:
```java
try {
  Replication replication = replicationFactory.newInstance(tableReplication);
  tableReplicationListener.tableReplicationStart(...);
  tableReplicationListener.tableReplicationSuccess(...);
} catch (Throwable t) {
  replicationFailures++;
  completionCode = CompletionCode.FAILURE;
  LOG.error("Failed to replicate: {}.", summary, t);
  tableReplicationListener.tableReplicationFailure(...);
```

If there's an issue with creating a new instance, then `startTime` might not have been initialised, but if everything works fine, `tableReplicationSuccess` should have the startTime that was set in `tableReplicationStart`.

So maybe this is fine?

massdosage: Yeah, that makes sense. There's quite a lot of assumptions made in this class based on the order that things are called on it which isn't ideal but that's a different topic ;)

When the above failure happens is the listener ever called again? i.e. could the time get set in the failure, and it then tries to do something else which also fails, at which point the time in the failure would be the previous failure's start time?

circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
AnanaMJ: Done :)

circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListener.java
massdosage: Yes, I'm fine with leaving the assumptions on call order for another day. I just wonder if it's worth clearing the startTime though?

patduin: that's true, that's but in that case the whole start-end time is rather useless info anyway.
Pretty sure the null check is there to have a quick nullPointerException fix and have "some" value in the failure message.

patduin: I mean by all means set startTime to null after a success, just in case. That won't hurt.

circus-train-aws-sns/src/test/java/com/hotels/bdp/circustrain/aws/sns/event/SnsListenerTest.java
```diff
@@ -322,4 +325,182 @@ public void messageSizeExceeded() throws Exception {
...
+            + "\"replicaTableLocation\":\""
+            + REPLICA_TABLE_LOCATION
+            + "\",\"replicaMetastoreUris\":\""
+            + REPLICA_METASTORE_URIS
+            + "\",\"bytesReplicated\":40}"));
+  }
+
+  @Test
+  public void noTableReplicationStartForConsecutiveReplications() {
+    SnsListener listener = new SnsListener(client, config, clock);
```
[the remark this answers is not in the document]
patduin: this is what I meant with calling it twice, all good

PR #116, opened 2019-02-01
only schedule table locations of unpartitioned tables

patduin:


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -98,14 +98,21 @@ public void updateMetadata(
     try (CloseableMetaStoreClient client = getMetaStoreClientSupplier().get()) {
       Optional<Table> oldReplicaTable = updateTableMetadata(client, eventId, sourceTable, replicaDatabaseName,
           replicaTableName, locationManager.getTableLocation(), replicationMode);
-      if (oldReplicaTable.isPresent() && LocationUtils.hasLocation(oldReplicaTable.get())) {
+      if (oldReplicaTable.isPresent()
+          && LocationUtils.hasLocation(oldReplicaTable.get())
+          && isUnpartitioned(oldReplicaTable.get())) {
```
patduin: I opted for this fix so we'll never add the location of an unpartitioned table.

Alternatively we could change the call, to call the updateMetadata with the partitions argument and make sure that works ok if partitions is an empty list. I've opted for above fix as I think any accidental call to this method should never schedule the table location for a partitioned table.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
```diff
@@ -163,9 +163,6 @@ public void prepare() throws Exception {
   private void convertSourceTableToView() {
     sourceTable.setTableType(TableType.VIRTUAL_VIEW.name());
     sourceTable.getSd().setLocation(null);
-    when(mockSourceLocationManager.getTableLocation()).thenReturn(null);
-    when(mockSourceLocationManager.getPartitionLocations()).thenReturn(Collections.<Path> emptyList());
-    when(mockSourceLocationManager.getPartitionSubPath(any(Path.class))).thenReturn(null);
```
AnanaMJ: `private @Mock SourceLocationManager mockSourceLocationManager;` can also be removed

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
```diff
@@ -515,6 +518,15 @@ public int compare(ColumnStatistics o1, ColumnStatistics o2) {
     assertThat(columnStats.get(1).getStatsObj().size(), is(2));
   }
 
+  @Test
+  public void updateMetadataCalledWithoutPartitionsDoesCleanUpLocations() throws TException, IOException {
```
massdosage: Shouldn't this be "doesnt" clean up locations? not "does"?

This is testing what happens with a partitioned table where no partitions have been altered right? So the table location doesn't get scheduled.

patduin: yes, well spotted!

Issue #56, opened 2018-04-09
s3s3Copier fails for large data sets


yashrajrs:
s3S3Copier fails for large data sets for both partitioned as well as unpartitioned tables.

```
com.hotels.bdp.circustrain.api.CircusTrainException: Unable to replicate
	at com.hotels.bdp.circustrain.core.PartitionedTableReplication.replicate(PartitionedTableReplication.java:136)
	at com.hotels.bdp.circustrain.core.Locomotive.run(Locomotive.java:114)
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:791)
	at org.springframework.boot.SpringApplication.callRunners(SpringApplication.java:781)
	at org.springframework.boot.SpringApplication.afterRefresh(SpringApplication.java:771)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:307)
	at com.hotels.bdp.circustrain.CircusTrain.main(CircusTrain.java:87)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.apache.hadoop.util.RunJar.run(RunJar.java:221)
	at org.apache.hadoop.util.RunJar.main(RunJar.java:136)
Caused by: com.hotels.bdp.circustrain.api.CircusTrainException: Error in S3S3Copier:
	at com.hotels.bdp.circustrain.s3s3copier.S3S3Copier.copy(S3S3Copier.java:101)
	at com.hotels.bdp.circustrain.core.PartitionedTableReplication.replicate(PartitionedTableReplication.java:121)
	... 12 more
Caused by: com.hotels.shaded.com.amazonaws.services.s3.model.AmazonS3Exception: We encountered an internal error. Please try again. (Service: Amazon S3; Status Code: 200; Error Code: InternalError; Request ID: 7EE3EE544247CBB2), S3 Extended Request ID: 7CbLnoxgeTt3+wfLuj6yckKDIxyUo2jEH4LMKDwE77qx5d8jjWYY9WHbIJpzHgcZ71OH4l2PnjY=
[...]

PR #119, opened 2019-02-18
Add Spring retry to S3S3copier

max-jacobs:
Fix for [issue 56](https://github.com/HotelsDotCom/circus-train/issues/56).

`RetrayableTransferManager` wraps `TransferManager.copy()` in order to provide a retry mechanism, implemented using Spring retry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## 13.2.2 - 2019-02-20
-### Changed
-* Added a retry mechanism to handle AWS S3 copying bug
-* Added a configuration parameter to support a configurable number of maximum attempts at copying
+### Added
+* Configurable retry mechanism to handle flaky AWS S3 to S3 copying. See [#56](https://github.com/HotelsDotCom/circus-train/issues/56).
 
 ## TBD
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 13.2.2 - 2019-02-20
+### Changed
+* Added a retry mechanism to handle AWS S3 copying bug
+* Added a configuration parameter to support a configurable number of maximum attempts at copying
+
 ## TBD
 ### Fixed
 * Clear partitioned state correctly for `SnsListener`. See [#104](https://github.com/HotelsDotCom/circus-train/issues/104).
--- a/circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainS3S3IntegrationTest/partitioned-single-table-with-no-partitions-mirror.yml
+++ b/circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainS3S3IntegrationTest/partitioned-single-table-with-no-partitions-mirror.yml
@@ -8,3 +8,5 @@ table-replications:
       table-name: ct_table_p_copy
 security:
   credential-provider: jceks://file/${config-location}/aws.jceks
+copier-options:
+  s3s3copier-retry-maxattempts: 3
--- a/circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
+++ b/circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
@@ -29,7 +29,6 @@ import org.slf4j.LoggerFactory;
 import com.amazonaws.AmazonClientException;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3URI;
-import com.amazonaws.services.s3.model.AmazonS3Exception;
 import com.amazonaws.services.s3.model.CopyObjectRequest;
 import com.amazonaws.services.s3.model.ListObjectsRequest;
 import com.amazonaws.services.s3.model.ObjectListing;
@@ -38,7 +37,6 @@ import com.amazonaws.services.s3.model.S3ObjectSummary;
 import com.amazonaws.services.s3.transfer.Copy;
 import com.amazonaws.services.s3.transfer.Transfer;
 import com.amazonaws.services.s3.transfer.Transfer.TransferState;
-import com.amazonaws.services.s3.transfer.TransferManager;
 import com.amazonaws.services.s3.transfer.internal.TransferStateChangeListener;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
@@ -50,12 +48,12 @@ import com.hotels.bdp.circustrain.api.copier.Copier;
 import com.hotels.bdp.circustrain.api.metrics.Metrics;
 import com.hotels.bdp.circustrain.s3s3copier.aws.AmazonS3ClientFactory;
 import com.hotels.bdp.circustrain.s3s3copier.aws.ListObjectsRequestFactory;
-import com.hotels.bdp.circustrain.s3s3copier.aws.TransferManagerFactory;
+import com.hotels.bdp.circustrain.s3s3copier.aws.RetryableTransferManager;
+import com.hotels.bdp.circustrain.s3s3copier.aws.RetryableTransferManagerFactory;
 
 public class S3S3Copier implements Copier {
 
   private static final Logger LOG = LoggerFactory.getLogger(S3S3Copier.class);
-  private static final int MAX_COPY_ATTEMPTS = 2;
 
   private static class BytesTransferStateChangeListener implements TransferStateChangeListener {
 
@@ -106,10 +104,10 @@ public class S3S3Copier implements Copier {
   private final ListObjectsRequestFactory listObjectsRequestFactory;
   private final MetricRegistry registry;
   private final AmazonS3ClientFactory s3ClientFactory;
-  private final TransferManagerFactory transferManagerFactory;
+  private final RetryableTransferManagerFactory retryableTransferManagerFactory;
   private final S3S3CopierOptions s3s3CopierOptions;
 
-  private TransferManager transferManager;
+  private RetryableTransferManager retryableTransferManager;
   private final List<Copy> copyJobs = new ArrayList<>();
 
   private long totalBytesToReplicate = 0;
@@ -122,7 +120,7 @@ public class S3S3Copier implements Copier {
       List<Path> sourceSubLocations,
       Path replicaLocation,
       AmazonS3ClientFactory s3ClientFactory,
-      TransferManagerFactory transferManagerFactory,
+      RetryableTransferManagerFactory retryableTransferManagerFactory,
       ListObjectsRequestFactory listObjectsRequestFactory,
       MetricRegistry registry,
       S3S3CopierOptions s3s3CopierOptions) {
@@ -130,7 +128,7 @@ public class S3S3Copier implements Copier {
     this.sourceSubLocations = sourceSubLocations;
     this.replicaLocation = replicaLocation;
     this.s3ClientFactory = s3ClientFactory;
-    this.transferManagerFactory = transferManagerFactory;
+    this.retryableTransferManagerFactory = retryableTransferManagerFactory;
     this.listObjectsRequestFactory = listObjectsRequestFactory;
     this.registry = registry;
     this.s3s3CopierOptions = s3s3CopierOptions;
@@ -147,8 +145,8 @@ public class S3S3Copier implements Copier {
       }
     } finally {
       // cancel any running tasks
-      if (transferManager != null) {
-        transferManager.shutdownNow();
+      if (retryableTransferManager != null) {
+        retryableTransferManager.shutdownNow();
       }
     }
   }
@@ -158,7 +156,7 @@ public class S3S3Copier implements Copier {
     AmazonS3URI targetBase = toAmazonS3URI(replicaLocation.toUri());
     srcClient = s3ClientFactory.newInstance(sourceBase, s3s3CopierOptions);
     targetClient = s3ClientFactory.newInstance(targetBase, s3s3CopierOptions);
-    transferManager = transferManagerFactory.newInstance(targetClient, s3s3CopierOptions);
+    retryableTransferManager = retryableTransferManagerFactory.newInstance(targetClient, s3s3CopierOptions, srcClient);
     if (sourceSubLocations.isEmpty()) {
       copy(sourceBase, targetBase);
     } else {
@@ -213,7 +211,7 @@ public class S3S3Copier implements Copier {
 
       TransferStateChangeListener stateChangeListener = new BytesTransferStateChangeListener(s3ObjectSummary,
           targetS3Uri, targetKey);
-      Copy copy = copyWithRetry(copyObjectRequest, stateChangeListener);
+      Copy copy = retryableTransferManager.copy(copyObjectRequest, stateChangeListener);
       totalBytesToReplicate += copy.getProgress().getTotalBytesToTransfer();
       copyJobs.add(copy);
     }
@@ -225,23 +223,6 @@ public class S3S3Copier implements Copier {
       objectMetadata.setSSEAlgorithm(ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
       copyObjectRequest.setNewObjectMetadata(objectMetadata);
     }
-  }
-
-  private Copy copyWithRetry(CopyObjectRequest copyObjectRequest, TransferStateChangeListener stateChangeListener) {
-    int copyAttempt = 1;
-    Copy copy = null;
-    while(copyAttempt <= MAX_COPY_ATTEMPTS) {
-      try {
-        copy = transferManager.copy(copyObjectRequest, srcClient, stateChangeListener);
-        break;
-      } catch (AmazonS3Exception e) {
-        if (++copyAttempt > MAX_COPY_ATTEMPTS) {
-          throw e;
-        }
-        LOG.error("S3 failed to copy object on attempt {}", copyAttempt, e);
-      }
-    }
-    return copy;
   }
 
   private Metrics gatherAllCopyResults() {
--- a/circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
+++ b/circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
@@ -29,6 +29,7 @@ import org.slf4j.LoggerFactory;
 import com.amazonaws.AmazonClientException;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3URI;
+import com.amazonaws.services.s3.model.AmazonS3Exception;
 import com.amazonaws.services.s3.model.CopyObjectRequest;
 import com.amazonaws.services.s3.model.ListObjectsRequest;
 import com.amazonaws.services.s3.model.ObjectListing;
@@ -53,7 +54,8 @@ import com.hotels.bdp.circustrain.s3s3copier.aws.TransferManagerFactory;
 
 public class S3S3Copier implements Copier {
 
-  private final static Logger LOG = LoggerFactory.getLogger(S3S3Copier.class);
+  private static final Logger LOG = LoggerFactory.getLogger(S3S3Copier.class);
+  private static final int MAX_COPY_ATTEMPTS = 2;
 
   private static class BytesTransferStateChangeListener implements TransferStateChangeListener {
 
@@ -211,7 +213,7 @@ public class S3S3Copier implements Copier {
 
       TransferStateChangeListener stateChangeListener = new BytesTransferStateChangeListener(s3ObjectSummary,
           targetS3Uri, targetKey);
-      Copy copy = transferManager.copy(copyObjectRequest, srcClient, stateChangeListener);
+      Copy copy = copyWithRetry(copyObjectRequest, stateChangeListener);
       totalBytesToReplicate += copy.getProgress().getTotalBytesToTransfer();
       copyJobs.add(copy);
     }
@@ -223,6 +225,23 @@ public class S3S3Copier implements Copier {
       objectMetadata.setSSEAlgorithm(ObjectMetadata.AES_256_SERVER_SIDE_ENCRYPTION);
       copyObjectRequest.setNewObjectMetadata(objectMetadata);
     }
+  }
+
+  private Copy copyWithRetry(CopyObjectRequest copyObjectRequest, TransferStateChangeListener stateChangeListener) {
+    int copyAttempt = 1;
+    Copy copy = null;
+    while(copyAttempt <= MAX_COPY_ATTEMPTS) {
+      try {
+        copy = transferManager.copy(copyObjectRequest, srcClient, stateChangeListener);
+        break;
+      } catch (AmazonS3Exception e) {
+        if (++copyAttempt > MAX_COPY_ATTEMPTS) {
+          throw e;
+        }
+        LOG.error("S3 failed to copy object on attempt {}", copyAttempt, e);
+      }
+    }
+    return copy;
   }
 
   private Metrics gatherAllCopyResults() {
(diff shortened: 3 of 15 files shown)


circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
```diff
@@ -0,0 +1,54 @@
...
+  public RetryableTransferManager(TransferManager transferManager, AmazonS3 srcClient) {
+    this.transferManager = transferManager;
+    this.srcClient = srcClient;
+  }
+
+  @Retryable(value = { AmazonS3Exception.class },
+      backoff = @Backoff(delay = BACKOFF_DELAY_MS, multiplier = 2))
+  public Copy copy(CopyObjectRequest copyObjectRequest, TransferStateChangeListener stateChangeListener) {
+    retryAttempt++;
+    LOG.info("copying attempt {}/3", retryAttempt);
```
patduin: Where does this 3 come from, is that a spring default?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
```diff
@@ -0,0 +1,54 @@
...
+import com.amazonaws.services.s3.transfer.TransferManager;
+import com.amazonaws.services.s3.transfer.internal.TransferStateChangeListener;
+
+public class RetryableTransferManager {
+
+  private static final Logger LOG = LoggerFactory.getLogger(RetryableTransferManager.class);
+  private static final int BACKOFF_DELAY_MS = 500;
+  private TransferManager transferManager;
+  private AmazonS3 srcClient;
+  private int retryAttempt = 0;
```
patduin: should we start counting at 1 more human readable to read attempt 1/3, 2/3, 3/3 instead of 0/3, 1/3, 2/3?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
max-jacobs: Yep Spring defaults to 3 attempts. I've also noticed that if there are multiple objects being copied, the `retryAttempt` will continue to increment. Adding a test for this now so that the logs are correct.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
```diff
@@ -0,0 +1,54 @@
...
+    this.transferManager = transferManager;
+    this.srcClient = srcClient;
+  }
+
+  @Retryable(value = { AmazonS3Exception.class },
+      backoff = @Backoff(delay = BACKOFF_DELAY_MS, multiplier = 2))
+  public Copy copy(CopyObjectRequest copyObjectRequest, TransferStateChangeListener stateChangeListener) {
+    retryAttempt++;
+    LOG.info("copying attempt {}/3", retryAttempt);
+    return transferManager.copy(copyObjectRequest, srcClient, stateChangeListener);
```
patduin: is it ok to reuse the client on every attempt or should we create a new one?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
```diff
@@ -0,0 +1,54 @@
...
+  private int retryAttempt = 0;
+
+  public RetryableTransferManager(TransferManager transferManager, AmazonS3 srcClient) {
+    this.transferManager = transferManager;
+    this.srcClient = srcClient;
+  }
+
+  @Retryable(value = { AmazonS3Exception.class },
+      backoff = @Backoff(delay = BACKOFF_DELAY_MS, multiplier = 2))
+  public Copy copy(CopyObjectRequest copyObjectRequest, TransferStateChangeListener stateChangeListener) {
```
patduin: Can we configure the number of attempts to try??

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
patduin: ok, let's make the nr of attempts configurable. I guess we can't use the spring property as that would be the same for all RetryTemplates (we'd have a clash if we ever add another in the future??).

max-jacobs: Do you mean make it configurable for the user? If so I will have to refactor as annotation values must be constants and so aren't configurable.

And yes, with this setup, the spring property would apply across all retries. It may be possible to maintain multiple `RetryTemplate`s to avoid having clashes.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
max-jacobs: The logging count does indeed start at 1, the increment occurs just before the log each time :)

patduin: ah of course my bad

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
max-jacobs: This is configured in the same way as the current setup. The client is instantiated at the very beginning when the copy jobs are started and doesn't change throughout this process. Any new calls to the top level `copy()` method in `S3S3Copier` will create a new client.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
abhimanyugupta07: Also, it would be nice to declare '3' as a constant (perhaps, MAX_RETRIES).

max-jacobs: @abhimanyugupta07 I've gone with `MAX_ATTEMPTS` as it includes the first failure :)

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -156,7 +156,7 @@ private void startAllCopyJobs() {
     AmazonS3URI targetBase = toAmazonS3URI(replicaLocation.toUri());
     srcClient = s3ClientFactory.newInstance(sourceBase, s3s3CopierOptions);
     targetClient = s3ClientFactory.newInstance(targetBase, s3s3CopierOptions);
-    transferManager = transferManagerFactory.newInstance(targetClient, s3s3CopierOptions);
+    retryableTransferManager = retryableTransferManagerFactory.newInstance(targetClient, s3s3CopierOptions, srcClient);
```
abhimanyugupta07: I see that `newInstance` is taking in `srcClient` as the argument now. Do we know why was it not like this in `TransferManagerFactory`. Just want to understand, if there was any specific reason for not tying up the transferManager to a sourceClient. If this coupling is unnecessary, than I think, it should be avoided.

@patduin thoughts ?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManagerFactory.java
```diff
@@ -24,15 +24,16 @@
 import com.hotels.bdp.circustrain.s3s3copier.S3S3CopierOptions;
 
 @Component
-public class TransferManagerFactory {
+public class RetryableTransferManagerFactory {
```
abhimanyugupta07: Why is this change necessary ? A transferManagerFactory is a Factory which returns different types of transferManagers, in your case now, you will return a RetryableTransferManager.

Please let me know if I am missing anything.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
abhimanyugupta07: And use that constant on line 49 in the log statment as well, please.

max-jacobs: ah yes, good spot, thanks!

patduin: this doesn't work: ``` maxAttempts="${s3toS3copier-retry-maxattempts}"``` with that value coming from the yml? Might have to use: `maxAttemptsExpression`

max-jacobs: It doesn't look like it unfortunately, Java annotations must be constants. Have I understood you correctly? 

As a workaround, I could hardcode the default number of attempts and then break from the retry early if the user specifies a lower number of attempts?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
patduin: ok cool, should be fine then!

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
patduin: it was already tied just went from `copy` see line 214 to the`newInstance` method. That's ok it is a field so makes more sense to load it in here. The copy method now has all arguments with the same scope.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
max-jacobs: Yep I have updated this now

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManagerFactory.java
max-jacobs: Hey, can I resolve this conversation? Thanks!

abhimanyugupta07: Yes Please.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
max-jacobs: This has now been refactored so that the `maxAttempts` property is configurable

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierOptions.java
```diff
@@ -54,7 +54,11 @@
      * {@link com.amazonaws.services.s3.model.CannedAccessControlList}
      * Optional AWS Canned ACL
      */
-    CANNED_ACL("canned-acl");
+    CANNED_ACL("canned-acl"),
+    /**
+     * Number of copy attempts to allow when copying from S3 to S3. Default value is 1.
```
patduin: Shall we just do 3 as a default? No harm right usually no retry is necesary

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManagerTest.java
```diff
@@ -0,0 +1,115 @@
...
+import com.amazonaws.services.s3.transfer.internal.TransferStateChangeListener;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(classes = RetryableTransferManagerTest.ContextConfiguration.class)
+public class RetryableTransferManagerTest {
+
+  private static int MAX_COPY_ATTEMPTS = 3;
+
+  @Autowired
+  private TransferManager mockedTransferManager;
```
patduin: What do we gain but injecting the classes via spring instead of mocking them directly?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierOptions.java
max-jacobs: Sure thing, that makes sense

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
```diff
@@ -0,0 +1,87 @@
...
+  public Copy copy(final CopyObjectRequest copyObjectRequest, final TransferStateChangeListener stateChangeListener) {
+    try {
+      return retryTemplate.execute(new RetryCallback<Copy, Throwable>() {
+        public Copy doWithRetry(RetryContext context) {
+          LOG.info("copying attempt {}/{}", context.getRetryCount()+1, maxAttempts);
+          return transferManager.copy(copyObjectRequest, srcClient, stateChangeListener);
+        }
+      });
+    } catch (Throwable throwable) {
+      throw new AmazonS3Exception(throwable.getMessage());
```
patduin: We should add the throwable as the cause otherwise we loose the original stacktrace

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManagerTest.java
max-jacobs: Yep these can be mocked directly, and the mockito runner can be used as well. I was playing around with autowiring spring properties but should have changed it back. Thanks!

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
max-jacobs: Good idea!

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManagerTest.java
patduin: 👍 always good to try out stuff!

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 13.2.2 - 2019-02-20
+### Changed
+* Added a retry mechanism to handle AWS S3 copying bug
+* Added a configuration parameter to support a configurable number of maximum attempts at copying
```
patduin: this line can go, enough we just mention the retry I think. Can you add a link to the issue see examples below

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 13.2.2 - 2019-02-20
+### Changed
```
abhimanyugupta07: I think this should be `Added` rather than `Changed`
and then you can remove the word 'Added' from the start of statement.

https://keepachangelog.com/en/1.0.0/

CHANGELOG.md
abhimanyugupta07: Or make it 
```Added a configurable retry mechanism to handle flaky AWS S3 to S3 copying```.

Comment (max-jacobs) [reply CHANGELOG.md]: Thanks both, will make this more concise

File: circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierOptionsTest.java [hunk 12]
```diff
@@ -96,4 +96,24 @@ public void getCannedAclDefaultIsNull() throws Exception {
...
 
+  @Test
+  public void getMaxCopyAttempts() throws Exception {
+    copierOptions.put(S3S3CopierOptions.Keys.MAX_COPY_ATTEMPTS.keyName(), 3);
+    S3S3CopierOptions options = new S3S3CopierOptions(copierOptions);
+    assertThat(options.getMaxCopyAttempts(), is(3));
+  }
+
+  @Test
+  public void getMaxCopyAttemptsDefaultIsOne() throws Exception {
```
abhimanyugupta07: When we change the defaults to Three, please rename this test.

CHANGELOG.md
max-jacobs: Thanks for the link, looks interesting. Will update now.

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierOptionsTest.java
```diff
@@ -96,4 +96,24 @@ public void getCannedAclDefaultIsNull() throws Exception {
...
+    S3S3CopierOptions options = new S3S3CopierOptions(copierOptions);
+    assertThat(options.getMaxCopyAttempts(), is(3));
+  }
+
+  @Test
+  public void getMaxCopyAttemptsDefaultIsOneIfLessThanOne() throws Exception {
```
abhimanyugupta07: This one needs renaming too.

abhimanyugupta07: Shouldn't there be a validation check at the config level that the numAttempts can't be less than one ?

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainS3S3IntegrationTest/partitioned-single-table-with-no-partitions-mirror.yml
```diff
@@ -8,3 +8,5 @@ table-replications:
       table-name: ct_table_p_copy
 security:
   credential-provider: jceks://file/${config-location}/aws.jceks
+copier-options:
+  s3s3copier-retry-maxattempts: 3
```
massdosage: We might want to consider renaming this to s3-copy-retry-maxattempts so we can use the same option for both the S3S3Copier and the S3MapReduceCopier (if we decide to add a similar retry feature to the latter).

massdosage: Or if we don't add it to the S3MapReduceCopier we should rename this to `s3s3-retry-max-attempts` to be more consistent with the other s3s3 specific options (e.g. `s3s3-multipart-copy-threshold-in-bytes`, `s3s3-multipart-copy-part-size-in-bytes`).

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
```diff
@@ -0,0 +1,88 @@
...
+import com.amazonaws.services.s3.model.AmazonS3Exception;
+import com.amazonaws.services.s3.model.CopyObjectRequest;
+import com.amazonaws.services.s3.transfer.Copy;
+import com.amazonaws.services.s3.transfer.TransferManager;
+import com.amazonaws.services.s3.transfer.internal.TransferStateChangeListener;
+
+public class RetryableTransferManager {
+
+  private static final Logger LOG = LoggerFactory.getLogger(RetryableTransferManager.class);
+  private static final int INITIAL_RETRY_INTERVAL_MS = 500;
```
massdosage: Let's go for slightly higher values here, it's not entirely clear what the error is on the AWS side but I have a feeling we should wait a bit longer for it to resolve itself. So let's go for 2000 for initial and 6000 for max.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
```diff
@@ -0,0 +1,88 @@
...
+    this.maxAttempts = maxAttempts;
+    this.retryTemplate = setUpRetryTemplate();
+  }
+
+  public Copy copy(final CopyObjectRequest copyObjectRequest, final TransferStateChangeListener stateChangeListener) {
+    try {
+      return retryTemplate.execute(new RetryCallback<Copy, Throwable>() {
+        public Copy doWithRetry(RetryContext context) {
+          LOG.info("copying attempt {}/{}", context.getRetryCount()+1, maxAttempts);
+          return transferManager.copy(copyObjectRequest, srcClient, stateChangeListener);
```
massdosage: As discussed, the error might not actually be throw here it might be later on when we go through the list of submitted copy jobs and wait for them to complete.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainS3S3IntegrationTest/partitioned-single-table-with-no-partitions-mirror.yml
max-jacobs: Seeing as the map reduce copier has a retry already built in, shall we go with `s3s3-retry-max-copy-attempts` for consistency with the other options for now?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
max-jacobs: Hmm, in this case I can add some logging so we can capture this more clearly.

The better solution would be to refactor the retry logic so that we can retry the job if it fails while it is completing - either by introducing some parallelism, or by caching the jobs as they are submitted and retrying if they are discovered to have failed when we inspect their progress.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainS3S3IntegrationTest/partitioned-single-table-with-no-partitions-mirror.yml
massdosage: Yes.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/RetryableTransferManager.java
massdosage: Agreed, let's see if AWS respond to the case we have open with them for the underlying issue on their side and we can then decide on next steps.

Issue #102, opened 2018-12-14
Upgrade AWS Java SDK to latest 1.x version (1.11.469) and fix AWS deprecated warning


massdosage:
**Acceptance Criteria**

* AWS Java SDK dependency upgraded to to 1.11.469 (or whatever the current latest 1.x version is when this is done). 
* The following deprecated warnings removed
> [WARNING] /home/awoodhead/workspace/circus-train-parent/circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsConfiguration.java:[45,35] AmazonSNSAsyncClient(com.amazonaws.auth.AWSCredentialsProvider,com.amazonaws.ClientConfiguration,java.util.concurrent.ExecutorService) in com.amazonaws.services.sns.AmazonSNSAsyncClient has been deprecated
> [WARNING] /home/awoodhead/workspace/circus-train-parent/circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsConfiguration.java:[48,11] setRegion(com.amazonaws.regions.Region) in com.amazonaws.AmazonWebServiceClient has been deprecated

PR #121, opened 2019-02-25
Upgraded aws sdk to remove deprecation warning

max-jacobs:
Fixes [issue 102](https://github.com/HotelsDotCom/circus-train/issues/102).

AWS has deprecated the constructors of `AmazonSNSAsyncClient` in preference for a builder pattern. The builder returns an async client which no longer exposes the underlying executor, `AmazonSNSAsync`. This means that `awaitTermination()` can no longer be used for a graceful shutdown. However, this seems to be okay as the client is only used to publish synchronously thus the executor isn't used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * Clear partitioned state correctly for `SnsListener`. See [#104](https://github.com/HotelsDotCom/circus-train/issues/104).
 * Fixed issue where in certain cases the table location of a partitioned table would be scheduled for housekeeping.
 * Removed default script for creating a housekeeping schema to allow the use of schemas that are already created. See [#111](https://github.com/HotelsDotCom/circus-train/issues/111).
-* Upgraded aws sdk to remove deprecation warning. See [#102](https://github.com/HotelsDotCom/circus-train/issues/102).
+* Upgraded AWS SDK to remove deprecation warning. See [#102](https://github.com/HotelsDotCom/circus-train/issues/102).
 
 ### Added
 * Configurable retry mechanism to handle flaky AWS S3 to S3 copying. See [#56](https://github.com/HotelsDotCom/circus-train/issues/56).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,3 @@
-## 13.2.2 - 2019-02-25
-### Fixed
-* Upgraded aws sdk to remove deprecation warning. See [#102](https://github.com/HotelsDotCom/circus-train/issues/102).
-
-## 13.2.2 - 2019-02-20
-### Added
-* Configurable retry mechanism to handle flaky AWS S3 to S3 copying. See [#56](https://github.com/HotelsDotCom/circus-train/issues/56).
-
 ## TBD
 ### Changed
 * Updated `housekeeping` version to 3.0.6 (was 3.0.5).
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -50,7 +50,7 @@
     <shade.prefix>${project.groupId}.shaded</shade.prefix>
     <spring-platform.version>2.0.8.RELEASE</spring-platform.version>
     <!-- BEGIN: AWS jdk version + dependencies -->
-    <aws-jdk.version>1.11.431</aws-jdk.version>
+    <aws-jdk.version>1.11.503</aws-jdk.version>
```
abhimanyugupta07: They released another one 3 days back: 1.11.505. You might as well upgrade to that one.

https://github.com/aws/aws-sdk-java/releases

CHANGELOG.md
```diff
@@ -11,6 +7,10 @@
 * Clear partitioned state correctly for `SnsListener`. See [#104](https://github.com/HotelsDotCom/circus-train/issues/104).
 * Fixed issue where in certain cases the table location of a partitioned table would be scheduled for housekeeping.
 * Removed default script for creating a housekeeping schema to allow the use of schemas that are already created. See [#111](https://github.com/HotelsDotCom/circus-train/issues/111).
+* Upgraded aws sdk to remove deprecation warning. See [#102](https://github.com/HotelsDotCom/circus-train/issues/102).
```
abhimanyugupta07: aws -> 'AWS'

circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsConfiguration.java
```diff
@@ -32,21 +33,30 @@
...
-  AmazonSNSAsyncClient amazonSNS(ListenerConfig config, AWSCredentialsProvider awsCredentialsProvider) {
-    AmazonSNSAsyncClient client = new AmazonSNSAsyncClient(awsCredentialsProvider,
-        new ClientConfiguration().withRetryPolicy(PredefinedRetryPolicies.getDefaultRetryPolicy()),
-        Executors.newSingleThreadScheduledExecutor());
-    client.setRegion(Region.getRegion(Regions.fromName(config.getRegion())));
-    return client;
+  AmazonSNSAsync amazonSNS(ListenerConfig config, AWSCredentialsProvider awsCredentialsProvider) {
+    ClientConfiguration clientConfiguration = new ClientConfiguration().withRetryPolicy(
+        PredefinedRetryPolicies.getDefaultRetryPolicy());
+    ExecutorFactory executorFactory = new ExecutorFactory() {
```
patduin: what's the default executorFactory, perhaps we don't need to override this at all

CHANGELOG.md
massdosage: sdk -> SDK

circus-train-aws-sns/src/main/java/com/hotels/bdp/circustrain/aws/sns/event/SnsConfiguration.java
massdosage: Same for the retry policy.

max-jacobs: Thanks for the comments.

If not provided, the retry will indeed default to the same configuration that we are providing. So I will remove this.

The executory will default to `newFixedThreadPool(maxConnections)` where `maxConnections` defaults to 50, so this is slightly different to `newSingleThreadScheduledExecutor()`. Shall I stick with our implementation or go with the default?

massdosage: Hmmmm, I see why we would want to leave it with a single threaded one as that won't change any behaviour. On the other hand unless there's a really good reason to deviate from the defaults we shouldn't (convention over configuration). It's unclear here why we would deviate. Thoughts from anyone else?

patduin: I'm ok either setting, I suspect it was done to prevent waste. We are just sending a handful of sns messages. Premature optimisation maybe I have no idea.

max-jacobs: If we're happy to go with the default, I'll go with that instead to stick with convention, thanks both!

PR #126, opened 2019-02-28
Added retry mechanism when waiting for copy job to complete

max-jacobs:
Fixes [issue #56](https://github.com/HotelsDotCom/circus-train/issues/56).

The original retry mechanism, added in in [pull request #119](https://github.com/HotelsDotCom/circus-train/pull/119), retries if S3 throws an error when a copy job is submitted. It has now been verified that the bug raised in [issue #56](https://github.com/HotelsDotCom/circus-train/issues/56) occurs during `Copy.waitForCompletion()`, and not when the jobs are submitted.

The changes here remove the `RetryableTransferManager`, and replaces it with a retry mechanism which catches failed jobs, re-submits them, and re-gathers the results. The original configuration option is retained.


circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -225,22 +226,74 @@ private void applyObjectMetadata(CopyObjectRequest copyObjectRequest) {
...
+                .info("Replicating...': {}% complete",
+                    String.format("%.0f", (alreadyReplicated / (double) totalBytesToReplicate) * 100.0));
+          }
+        } catch (AmazonClientException e) {
           LOG
-              .info("Replicating...': {}% complete",
-                  String.format("%.0f", (alreadyReplicated / (double) totalBytesToReplicate) * 100.0));
+              .info("Copying '{}/{}' failed, adding to retry list.",
+                  copyJob.getCopyObjectRequest().getSourceBucketName(), copyJob.getCopyObjectRequest().getSourceBucketName());
+          failedCopyJobs.add(copyJob);
```
max-jacobs: An overhead with collecting failed jobs in this way is that a long running job will prevent subsequent jobs, which have already failed, from being resubmitted immediately. Is this an issue?

massdosage: I think we can live with this.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -225,22 +226,74 @@ private void applyObjectMetadata(CopyObjectRequest copyObjectRequest) {
...
+
+  private Copy submitCopyJob(CopyJob copyJob) {
+    CopyObjectRequest copyObjectRequest = copyJob.getCopyObjectRequest();
+    LOG
+        .info("Copying object from '{}/{}' to '{}/{}'", copyObjectRequest.getSourceBucketName(), copyObjectRequest.getSourceKey(),
+            copyObjectRequest.getDestinationBucketName(), copyObjectRequest.getDestinationKey());
+    return transferManager.copy(copyObjectRequest, srcClient, copyJob.getTransferStateChangeListener());
+  }
+
+  private List<CopyJob> gatherCopyJobs(List<CopyJob> copyJobs) {
```
massdosage: probably worth adding javadoc here just to say that the returned `List` contains the failed jobs

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierTest.java
```diff
@@ -402,25 +404,97 @@ public void copySafelyShutDownTransferWhenRetryFails() throws Exception {
...
+      ArgumentCaptor<CopyObjectRequest> captor = ArgumentCaptor.forClass(CopyObjectRequest.class);
+      verify(mockedTransferManager, Mockito.times(3)).copy(captor.capture(), any(AmazonS3.class), any(TransferStateChangeListener.class));
+      List<CopyObjectRequest> capturedCopyRequests = captor.getAllValues();
+      assertThat(capturedCopyRequests.get(0).getSourceKey(), is(sourceKey1));
+      assertThat(capturedCopyRequests.get(1).getSourceKey(), is(sourceKey2));
+      assertThat(capturedCopyRequests.get(2).getSourceKey(), is(sourceKey1));
+      verify(mockedTransferManager).shutdownNow();
+      verifyNoMoreInteractions(mockedTransferManager);
+    } catch (CircusTrainException e) {
+      fail("Exception should not have been thrown");
```
massdosage: @patduin feel free to chip in here, but we could remove the try catch and let this fail the test if thrown without having `fail` - at least that way we would see the stack trace of `e` which would give us a better idea of why it failed.

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierTest.java
```diff
@@ -402,25 +404,97 @@ public void copySafelyShutDownTransferWhenRetryFails() throws Exception {
...
+        any(TransferStateChangeListener.class))).thenReturn(copy);
+    TransferProgress transferProgress = new TransferProgress();
+    transferProgress.setTotalBytesToTransfer(7);
+    when(copy.getProgress()).thenReturn(transferProgress);
+    doThrow(new AmazonClientException("cause"))
+        .doNothing().when(copy).waitForCompletion();
+    S3S3Copier s3s3Copier = new S3S3Copier(sourceBaseLocation, sourceSubLocations, replicaLocation, s3ClientFactory,
+        mockedTransferManagerFactory, listObjectsRequestFactory, registry, s3S3CopierOptions);
+    try {
+      Metrics metrics = s3s3Copier.copy();
```
massdosage: Given the amount of boilerplate above I think i'd prefer merging these assertions into the above.

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierTest.java
```diff
@@ -291,23 +291,23 @@ public void copyCheckTransferManagerIsShutdownWhenCopyExceptionsAreThrown() thro
...
-        mockedRetryableTransferManagerFactory, listObjectsRequestFactory, registry, s3S3CopierOptions);
+        mockedTransferManagerFactory, listObjectsRequestFactory, registry, s3S3CopierOptions);
     try {
       s3s3Copier.copy();
       fail("exception should have been thrown");
     } catch (CircusTrainException e) {
-      verify(mockedRetryableTransferManager).shutdownNow();
-      assertThat(e.getCause().getMessage(), is("cause"));
+      verify(mockedTransferManager).shutdownNow();
+      assertThat(e.getMessage(), is("1 job(s) failed the maximum number of copy attempts, 3"));
```
massdosage: We could check something like 

`verify(mockedTransferManager, Mockito.times(3)).copy(captor.capture(), any(AmazonS3.class), any(TransferStateChangeListener.class));`

here. Then possibly rename the method to be "maxRetriesExceeded".

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierTest.java
patduin: I think the intent of this test is to show that if an exception is thrown the exception is caught and the copy is retried and then everything is ok. If the exception is bubbled up the retry logic has failed, which is what the tests nicely documents like this.
There are other tests where the copy succeeds and no retry logic is used, those will throw the exception you are after and appropriate stacktrace

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -225,22 +226,74 @@ private void applyObjectMetadata(CopyObjectRequest copyObjectRequest) {
...
+        Copy copy = copyJob.getCopy();
+        try {
+          copy.waitForCompletion();
+          long alreadyReplicated = bytesReplicated.addAndGet(copy.getProgress().getTotalBytesToTransfer());
+          if (totalBytesToReplicate > 0) {
+            LOG
+                .info("Replicating...': {}% complete",
+                    String.format("%.0f", (alreadyReplicated / (double) totalBytesToReplicate) * 100.0));
+          }
+        } catch (AmazonClientException e) {
```
patduin: We're not logging this anywhere now. Perhaps add a `LOG.debug("Copying failed, retrying, exception is", e); ` or something similar.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/CopyJob.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package com.hotels.bdp.circustrain.s3s3copier;
+
+import com.amazonaws.services.s3.model.CopyObjectRequest;
+import com.amazonaws.services.s3.transfer.Copy;
+import com.amazonaws.services.s3.transfer.internal.TransferStateChangeListener;
+
+public class CopyJob {
+  private Copy copy;
+  private CopyObjectRequest copyObjectRequest;
```
patduin: I think these two fields need to be their own object: `CopyJobRequest`.  Then you can get have just one constructor `CopyJob(copy, copyJobRequest)`. I think that will look nicer in the S3Copier, first create CopyJobRequests then from those create CopyJobs.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
patduin: or have a `GatherCopyJobResult` object that has `getFailedCopyJobs` method. (I am also fine with a line of javadoc.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -194,12 +200,9 @@ private void submitCopyJobsFromListing(
             request.getPrefix());
     List<S3ObjectSummary> objectSummaries = listing.getObjectSummaries();
     for (final S3ObjectSummary s3ObjectSummary : objectSummaries) {
+      totalBytesToReplicate += s3ObjectSummary.getSize();
```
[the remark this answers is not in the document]
max-jacobs: This is the final assert done in `copyGatherCorrectMetricsWhenJobsAreRetried()`, is this what you mean?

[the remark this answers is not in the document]
massdosage: More like writing a temporary debug statement to output the value and then run it for real once to check. The tests may not be correct as they use a proxy, not the real thing...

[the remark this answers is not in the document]
max-jacobs: Ah yes, I remember we discussed that, cheers

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/CopyJob.java
max-jacobs: Thanks, that does clean it up a little. Now there no need for `CopyJob`s with a null field.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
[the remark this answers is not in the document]
max-jacobs: Tested the real thing and it works as expected

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
max-jacobs: Added a Javadoc to the method

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
max-jacobs: Added an extra debug line to log the exception

PR #128, opened 2019-04-30
Improve readme

massdosage:
I've made the following changes which hopefully are an improvement:
* Full feature list (copied over from our internal blog) as I don't think we've been "selling" all the amazing things Circus Train can do very well here.
* Added links to related open source projects (this was the reason I started looking at the README).
* Changed the installation section to link directly to .tgz and use the same text for "extract and set home dir" as we do in WaggleDance README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,41 @@ You can obtain Circus Train from Maven Central:
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.hotels/circus-train/badge.svg?subject=com.hotels:circus-train)](https://maven-badges.herokuapp.com/maven-central/com.hotels/circus-train) [![Build Status](https://travis-ci.org/HotelsDotCom/circus-train.svg?branch=master)](https://travis-ci.org/HotelsDotCom/circus-train) [![Coverage Status](https://coveralls.io/repos/github/HotelsDotCom/circus-train/badge.svg?branch=master)](https://coveralls.io/github/HotelsDotCom/circus-train?branch=master) ![GitHub license](https://img.shields.io/github/license/HotelsDotCom/circus-train.svg)
 
 ## Overview
-Circus Train replicates Hive tables between clusters on request. It replicates both the table's data and metadata. Unlike many other solutions it has a light touch, requiring no direct integration with Hive's core services. However, it is not event driven and does not know how tables differ between sites; it merely responds to requests to copy (meta-)data. It can copy either entire unpartitioned tables or user defined sets of partitions on partitioned tables. Circus Train employs snapshot isolation to minimise the impact of changing data at the source, and to allow consumers of data in the replica cluster to operate independently of ongoing replication tasks.
+Circus Train replicates Hive tables between clusters on request. It replicates both the table's data and metadata. Unlike many other solutions it has a light touch, requiring no direct integration with Hive's core services. It can copy either entire unpartitioned tables or user defined sets of partitions on partitioned tables. 
+
+Other features include:
+* Replication of any of the following:
+
+   * an entire table and all its data on each run.
+   * a sub-set of the data represented by a filter query (e.g. for a data set partitioned by load date you could run Circus Train once a day and replicate only the previous day's data).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,11 +6,37 @@ You can obtain Circus Train from Maven Central:
...
-
-## Install
-Download the version to use from [Maven Central](http://mvnrepository.com/artifact/com.hotels/circus-train/) and uncompress it in a directory of your choosing.
+Circus Train replicates Hive tables between clusters on request. It replicates both the table's data and metadata. Unlike many other solutions it has a light touch, requiring no direct integration with Hive's core services. It can copy either entire unpartitioned tables or user defined sets of partitions on partitioned tables. A more detailed overview and the background of this project can be found in this blog post: [Replicating big datasets in the cloud](https://medium.com/hotels-com-technology/replicating-big-datasets-in-the-cloud-c0db388f6ba2). 
+
+Other features include:
+* Replication of any of the following:
+   * an entire table and all its data on each run.
+   * a sub-set of the data represented by a filter query (e.g. for a data set partitioned by load date you could run Circus Train once a day and replicate only the previous day's data).
```
abhimanyugupta07: why not `subset` ?

massdosage: I blame whoever wrote that on the internal wiki I copied it off ;) (wasn't me I promise!). I'll change it.

README.md
```diff
@@ -6,11 +6,37 @@ You can obtain Circus Train from Maven Central:
...
+* Control of how the data is copied - what file attributes are copied, bandwidth throttling etc.
+* Data consistency checking is performed as part of the replication.
+* Numerous retry mechanisms are used in order to handle the many connectivity and reliability issues that can occur when replicating across locations.
+* Transport encryption for both metadata and data.
+* Use of a Java Keystore Hadoop Credentials Provider to keep authentication keys secure.
+* Connectivity with private networks through a bastion host and direct connections.
+* Table, partition, and column statistics are also copied to preserve performance at the replica.
+* Metrics about replication (success/failure, bandwidth, bytes copied etc.) can be sent to Graphite for building dashboards, alerting etc.
+* Pluggable replication event listeners so users can implement their own functionality based on events occurring during a replication (e.g. triggering a downstream job when a replication completes). A [sample listener](https://github.com/HotelsDotCom/circus-train/tree/master/circus-train-aws-sns) that uses Amazon's Simple Notification Service (SNS) is provided as a usable reference implementation of this.
```
abhimanyugupta07: A hyperlink over SNS would be great. https://aws.amazon.com/sns/

README.md
```diff
@@ -6,11 +6,37 @@ You can obtain Circus Train from Maven Central:
...
+* Pluggable transformations so users can implement their own functionality to transform table, partition and column statistics metadata during a replication.
```
abhimanyugupta07: Not a comment but more to understand, what does this feature do ? 

Feels like it should go above to the place where you first mentioned about statistics.

>Table, partition, and column statistics are also copied to preserve performance at the replica.

README.md
massdosage: Yeah, I didn't want this to turn into a link-fest, otherwise one could argue that above I should also hyperlink S3, Graphite, GCS etc. So for now I've just gone with linking to our listener, it's documentation should link to SNS somewhere (it currently doesn't but that's a different PR ;) )

README.md
massdosage: I guess the reason it's at the end is that the "pluggable" stuff was listed together. I think this is mainly talking about https://github.com/HotelsDotCom/circus-train/#metadata-transformations. I could hyperlink to that there and to https://github.com/HotelsDotCom/circus-train/#developing-a-replicationeventlistener for the feature above? Similar for Metrics and Housekeeping, might be nice, if someone is interested they can click here and see the more detailed docs on the feature.

README.md
abhimanyugupta07: Ok. I will keep this in mind :P

README.md
abhimanyugupta07: So, the feature is to add transformations for Table and partition `metadata` & column statistics.

And yeah, hyperlinking would be great.

massdosage: OK, added some hyperlinks to other relevant parts of the documentation

Issue #132, opened 2019-05-10
Don't override target cluster replication factor


patduin:
_As a user of CT_
_I'd like to change CT to use the target cluster replication factor by default_
_So I don't silently overwrite it_

CT uses the hadoop configuration (core-site.xml, etc..) on the cluster it runs on to configure the M/R job used for copying the data.  If you push data to a different cluster that means you might override the `dfs.replication` factor for that data using the source cluster setting and not the target cluster setting. 
We should change that behaviour and make sure we don't override that setting unless it is explicitly set in the CT yml configuration copier-options section.

It is particularly tricky when you run in EMR (dfs.replication=1 by default) and replicate to an on-premise HDFS cluster which usually has dfs.replication=3.

### Acceptance Criteria:
* Target cluster dfs.replication is not overridden unless explicitly configured in CT configuration.

patduin:
We can't really do this correctly, see comments in the PR. I'm closing this.

PR #133, opened 2019-05-29
Added some comments in the readme about 'dfs.replication'

patduin:
I've investigated #132 and since CT can run pushing data from a source cluster or pulling data from a target cluster it's very very hard to determine if whatever value is taken from the the Hadoop configuration is correct or should be overridden. I'm afraid this is one that a user needs to be aware of themselves. I've added a comment to warn for this and a clear example how to override the replication factor in Circus Train.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,15 @@ If you are planning to run Circus Train on EMR you will need to set up the EMR c
 
 Note that the paths above are correct as of when this document was last updated but may differ across EMR versions, refer to the [EMR release guide](http://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-release-components.html) for more up to date information if necessary.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -678,6 +678,15 @@ By default Circus Train uses an [H2](http://www.h2database.com) file-based datab
 ## Configuration Examples
 Below are a number of configuration fragments which can be used as a starting point for creating your own configuration files. These are for reference and will obviously need host names and various other values changed to suit your setup. Also keep in mind that database and table locations in Hive are fully qualified paths therefore if using AWS then EMR must be able to translate DNS names. If a DNS service is not available then each EMR node - master and slaves - must include an entry in _etc/hosts_ for each domain name in the Circus Train configuration file.
 
+*Note*: When replicating data to a Hadoop cluster you can set the replication factor. Circus Train uses the `dfs.replication` setting from the cluster that is running Circus Train. On EMR this replication factor is set to 1 by AWS defaults. The setting can be overridden by changing the `dfs.replication` value on the cluster that you run on. The setting can be found in `/etc/hadoop/conf/hdfs-site.xml`. Alternatively it can be set in the Circus Train configuration file like so:
```
massdosage: This is good but I think it would be better in the EMR section - https://github.com/HotelsDotCom/circus-train#emr

patduin: ok yeah sounds good

PR #135, opened 2019-06-28
Update template config in circus-train-package

AnanaMJ:


circus-train-package/src/main/resources/circus-train-minimal.yml.template
```diff
@@ -16,3 +16,4 @@ table-replications:
       table-name: <table>
     replica-table:
       table-location: s3://<company>-<team>-<database>-<table>/
+husekeeping.db-init-script: classpath:/schema.sql
```
max-jacobs: husekeeping -> housekeeping :)

AnanaMJ: 😱 
I blame vim

AnanaMJ: Done :)

PR #138, opened 2019-08-21
Integrate with beekeeper

max-jacobs:
This PR contains some of the changes required to integrate Circus Train with Beekeeper.

It include the addition of a new Table Transformation which will annotate a Hive table with parameters defined by the user in their configuration file. Each table replication can contain unique parameters which will override any default parameters.


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/BeekeeperHiveTableAnnotator.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.thrift.TException;
+
+import com.google.common.base.Supplier;
+
+import com.hotels.bdp.circustrain.api.CircusTrainException;
+import com.hotels.hcommon.hive.metastore.client.api.CloseableMetaStoreClient;
+
+public class BeekeeperHiveTableAnnotator implements HiveTableAnnotator {
```
patduin: I think this should be made generic. CT should probably not know about beekeeper at all. The values of these properties should come from the yam configuration.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/TableAnnotator.java
```diff
@@ -0,0 +1,21 @@
+package com.hotels.bdp.circustrain.core.annotation;
+
+import java.util.Map;
+
+import org.apache.hadoop.hive.metastore.api.Table;
+
+import com.hotels.bdp.circustrain.api.CircusTrainException;
+
+public interface TableAnnotator {
```
patduin: Not quite sure I understand why you need this and `HiveTableAnnotator`

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/TableAnnotatorFactory.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.hotels.bdp.circustrain.api.conf.OrphanedDataStrategy;
+import com.hotels.bdp.circustrain.api.conf.ReplicationMode;
+
+public class TableAnnotatorFactory {
+
+  public static TableAnnotator newInstance(ReplicationMode replicationMode,
+    OrphanedDataStrategy orphanedDataStrategy) {
+    if (replicationMode == ReplicationMode.FULL) {
+      if (orphanedDataStrategy == OrphanedDataStrategy.BEEKEEPER) {
```
patduin: I think if you want to use BEEKEEPER you need to set the OrphanedDataStrategy to NULL/NONE for CT and just add the table properties in a "generic" way.

max-jacobs: In this case we probably don't need `OrphanedDataStrategy` at all, just a housekeeping "on" or "off" option. Then `OrphanedDataOptions` is really just `ReplicaTableAnnotations` or something, and the user can just add whatever they like, including Beekeeper parameters which they can find in the docs.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/TableAnnotator.java
max-jacobs: There are two main paths, destructive and upsert. Destructive deletes the table without creating an intermediary table object (so the params would need to be added to the table in Hive directly), but upsert does create an intermediary java object, so I thought for the upsert path it would be easier to add params the params before it is then created in Hive.

Looking at it again perhaps I'm wrong, in the upsert path once the replica table is created in `Replica`, the params can be added using the `HiveTableAnnotator`. This would be just as easy.

Does that make sense?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/TableAnnotatorFactory.java
max-jacobs: I've updated this to three options for `OrphanedDataStrategy`: `HIVE_HOOK`, `HOUSEKEEPING` and `NONE` where `HIVE_HOOK` should add the options specified in `OrphanedDataOptions`.

What do people think about the name `HIVE_HOOK`? Trying to find a name that summarises the cleanup mechanism using hive events, but happy to change it.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/TableAnnotator.java
max-jacobs: Removed `TableAnnotator` now and just have the `HiveTableAnnotator`.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/TableAnnotatorFactory.java
patduin: Should not have it imo :). OrphanedDataOptions should be a more generic add table properties option. CT is just setting a table property it shouldn't have to know it sets a beekeeper property that's up to the user. HIVE_HOOK is the same as Beekeeper. From a user perspective if I want to use CT with beekeeper. I need to add the table property and turn off housekeeping. CT doesn't know about beekeeper nor about hive events.

max-jacobs: The thing is don't we only want to add table annotations (e.g. for Beekeeper) if it's in `FULL` replication mode - i.e. not just metadata update? In which case CT would need to know to only add table params when its in a specific housekeeping mode.

Or do we presume, like with Beekeeper, that metadata only updates are handled properly? In which case, yep we can just have `NONE` and `HOUSEKEEPING` and change `OrphanedDataOptions` to `ReplicaTableAnnotations` or something.

patduin: If that's the case, the user should remove the table parameters from the configuration. Beekeeper and CT are completely separate we can't really have any logic in CT that says oh if we are beekeeping let's do this or that. That would tie beekeeper and CT together and we won't be able to change one without affecting the other. The only thing driving beekeeping is having a property on the table, you can add that by hand or we provide a mechanism to add table properties in CT which is a bit nicer.

max-jacobs: As in, if the update is just a metadata update, at the moment we currently don't schedule a path for deletion in housekeeping. I thought that this would mean we also wouldn't want to emit any event with a beekeeper param.

But beekeeper handles metadata updates (by ignoring them), so maybe we dont care about this. We just want all events to be emitted with a table param that the user sets, and then whatever downstream cleaner picks it up can decide whether or not to do cleanup.

I think I follow you now, good plan. Will remove references to `HIVE_HOOK` and just have two strategy options. This makes things simpler :)

patduin: I guess if it is a metadata only update beekeeper will no location has  been changed so nothing to delete right.

max-jacobs:
This PR is now ready for review

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/HiveTableAnnotator.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public void annotateTable(String databaseName, String tableName, Map<String, String> parameters)
+    throws CircusTrainException {
+    if (parameters.isEmpty()) {
+      return;
+    }
+    try (CloseableMetaStoreClient client = replicaMetaStoreClientSupplier.get()) {
+      Table table = client.getTable(databaseName, tableName);
+      Map<String, String> tableParameters = table.getParameters();
```
patduin: Maybe: `new LinkedHashMap(table.getParameters())` Just in case the returned map is immutable. Linked to keep insertion order, not sure if that matters much but won't hurt.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/DestructiveReplication.java
```diff
@@ -52,6 +56,10 @@ public void replicate() throws CircusTrainException {
             + tableReplication.getQualifiedReplicaName()
             + "' is not controlled by circus train aborting replication, check configuration for correct replica name");
       }
+
+      hiveTableAnnotator.annotateTable(tableReplication.getReplicaDatabaseName(),
```
patduin: Not sure this is needed? I'd say we only have to annotate when we create a new table.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -81,12 +82,12 @@
...
       HousekeepingListener housekeepingListener,
       ReplicaCatalogListener replicaCatalogListener,
-      ReplicationMode replicationMode) {
+      TableReplication tableReplication) {
     super(replicaCatalog.getName(), replicaHiveConf, replicaMetaStoreClientSupplier);
     this.replicaCatalogListener = replicaCatalogListener;
     tableFactory = replicaTableFactory;
     this.housekeepingListener = housekeepingListener;
-    this.replicationMode = replicationMode;
+    this.tableReplication = tableReplication;
```
patduin: NOABD: maybe nicer of you keep the field and do `replicationMode = tableReplicaion.getReplicationMode()` in the constructor once.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/DestructiveReplication.java
max-jacobs: Yeah I'm not 100% sure about the annotation logic in this class. CT does `destructiveReplica.dropDeletedPartitions()` and `destructiveReplica.dropTable()` on the replica which are both then scheduled for housekeeping cleanup. In which case I thought it makes sense to annotate the table _before_ peforming these actions so the paths are cleaned up.

But now that the table annotations aren't strictly about housekeeping/beekeeper, I think you're right we should only be applying them when we create a new table.

I think this will mean that when a user runs a destructive replication, they will need to annotate their table manually first? (otherwise these paths will be missed)

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/annotation/HiveTableAnnotator.java
max-jacobs: Yep that makes sense, updated

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
max-jacobs: Agreed, done

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/conf/OrphanedDataStrategy.java
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright (C) 2016-2018 Expedia Inc.
```
massdosage: Can you run the Mycila plugin to get these all up to 2019 please?

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/destructive-unpartitioned-single-table-no-housekeeping-with-table-parameters.yml
```diff
@@ -0,0 +1,11 @@
...
+  - source-table:
+      database-name: ${circus-train-runner.database-name}
+      table-name: ct_table_u
+    replica-table:
+      table-location: ${circus-train-runner.replica-warehouse-uri}/${circus-train-runner.database-name}/ct_table_u
+      parameters:
+        table.parameter.first: first
+        table.parameter.second: second
+    orphaned-data-strategy: NONE
+    replication-strategy: PROPAGATE_DELETES
```
massdosage: Can you add line break at EOF please.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/partitioned-single-table-no-housekeeping-with-table-parameters.yml
```diff
@@ -0,0 +1,18 @@
...
+      parameters:
+        table.parameter.first: first
+        table.parameter.second: second
+    orphaned-data-strategy: NONE
+copier-options:
+  file-attribute: replication, blocksize, user, group, permission, checksumtype
+  preserve-raw-xattrs: false
+metrics-reporter:
+  period: 1
+  time-unit: SECONDS
```
massdosage: Ditto

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/DestructiveReplication.java
patduin: I guess there are 2 usecases:
1) new table is being replicated and we want to use beekeeper
2) existing table is already being replicated and we want to switch to beekeeper

To cater the first it's easy we just have to add the annotation to newly created tables.
For the second we could say the user needs tol add the annotation manually (I am leaning towards this). Or we go add something in CT that adds the annotation always when doing an update.

max-jacobs: I don't mind, depends on what the user would expect to happen/making sure we document it clearly. As it stands:

1. We are adding parameters to new and existing table replications
2. If we want to change this to only add the params on new table replications, the annotation here should be removed, and I will need to make a change `Replica` to only add the annotations if the table has been newly created.

I'll make the change so we can see what this looks like.

What do we all think?

max-jacobs:
Came across this class `com.hotels.bdp.circustrain.core.HiveEndpoint`, any opinions on whether adding table annotations as a public method in here makes more sense than creating the new class `HiveTableAnnotator`?

patduin:
> Came across this class `com.hotels.bdp.circustrain.core.HiveEndpoint`, any opinions on whether adding table annotations as a public method in here makes more sense than creating the new class `HiveTableAnnotator`?

new class please. the HiveEndpoint is used in some of the tools which don't deal with annotations

README.md
```diff
@@ -565,6 +573,33 @@ The configuration options for Graphite are:
...
+To use this feature, add properties to your configuration:
+
+        table-replications:
+           - ...
+        transform-options:
+          table-parameters:
+            first-parameter-override: value
+            second-parameter-override: value
+
+This will add `first.table.parameter` and `second.table.parameter` to the metadata of all tables in your replication. These options can be overridden for each table by adding configuration to a specific table replication:
```
patduin: `first.table.parameter` != `first-parameter-override`

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/TableParametersConfig.java
```diff
@@ -0,0 +1,38 @@
...
+package com.hotels.bdp.circustrain.core.conf;
+
+import java.util.Map;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties(prefix = "table-parameters")
+public class TableParametersConfig {
```
patduin: bit late to the party apologies for that but Hive refers to these things as table properties: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-listTableProperties
Maybe we use the same name will be easier to understand what it overrides.

README.md
max-jacobs: Good spot, thanks!

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/TableParametersConfig.java
max-jacobs: Yep in Hive they're properties, but on the Java Table object they're parameters. Maybe everything user facing should be properties, and internally parameters? Or do you think everything should just be properties? Thanks for the suggestion

patduin: oh that's a good point hmm tough one. What do the others think?

massdosage: Yeah, go Hive for once again forcing us to do horrible things! I'm good with using parameters internally and properties externally to match what Hive appear to do.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/transformation/TableParametersTransformationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    Table transformedTable = transformation.transform(table);
+    Map<String, String> tableParameters = transformedTable.getParameters();
+    assertThat(tableParameters.get(KEY), is(VALUE));
+  }
+
+  @Test
+  public void typicalImmutableTableParameters() {
+    table.setParameters(Collections.EMPTY_MAP);
+    Table transformedTable = transformation.transform(table);
+    Map<String, String> tableParameters = transformedTable.getParameters();
```
massdosage: Are there any other entries in this returned Map? If not perhaps you want to assert the size of it in this and all the other tests just to ensure there aren't any surprising things being added to it.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/partitioned-single-table-no-housekeeping-with-table-parameters.yml
```diff
@@ -0,0 +1,19 @@
...
+  - source-table:
+      database-name: ${circus-train-runner.database-name}
+      table-name: ct_table_p
+      partition-filter: (continent='Europe' AND country='UK') OR (continent='Asia' AND country='China')
+      partition-limit: 100
+    replica-table:
+      table-location: ${circus-train-runner.replica-warehouse-uri}/${circus-train-runner.database-name}/ct_table_p
+    transform-options:
+      table-parameters:
+        '[table.parameter.first]': first
```
massdosage: Is the need for the `'[` due to yml (or something else) to prevent it from doing unwanted things with those `.`s?

max-jacobs: They are, without the '[]' spring will read the property as a nested map, and not as a single String.

README.md
```diff
@@ -565,6 +573,33 @@ The configuration options for Graphite are:
 |`graphite.prefix`|No|e.g. `dev`|All metrics are prefixed with `prefix.namespace`. By default this is picked up from the `config` if provided.|
 | `graphite.namespace`|Yes|e.g. `com.company.team.circus-train.application`|All metrics are prefixed with `prefix.namespace`. Formatting of this is important for reporting from Graphite.|
 
+### Metadata transformations
+Circus Train can transform the metadata of a newly created replica table by adding table parameters before performing the replication. This is useful, for example, to ensure that events emitted by your Hive Metastore are tagged with specific properties.
```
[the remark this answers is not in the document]
max-jacobs: I guess it doesn't directly, I'm trying to convey the reason why we added this feature, might be useful for someone reading this section to know a potential use case for a metadata transformation.

Happy to remove if you think its irrelevant here

README.md
```diff
@@ -565,6 +573,33 @@ The configuration options for Graphite are:
...
+To use this feature, add properties to your configuration:
+
+        table-replications:
+           - ...
+        transform-options:
+          table-parameters:
+            first-parameter-override: value
+            second-parameter-override: value
+
+This will add `first-table-parameter` and `second-table-parameter` to the metadata of all tables in your replication. These options can be overridden for each replication:
```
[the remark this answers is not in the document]
max-jacobs: Thanks, that is clearer

README.md
```diff
@@ -565,6 +573,33 @@ The configuration options for Graphite are:
...
+           transform-options:
+             table-parameters:
+               first-parameter-override: override
+               second-parameter-override: override
+         transform-options:
+           table-parameters:
+             first-parameter-override: value
+             second-parameter-override: value
+
+To add properties to your table with non-alphanumeric characters (other than -), surround the key with single quotes and brackets like so: '[...]'.
```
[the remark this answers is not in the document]
max-jacobs: Done, thanks

README.md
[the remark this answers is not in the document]
max-jacobs: Added a line at the end of this section

README.md
```diff
@@ -565,6 +573,33 @@ The configuration options for Graphite are:
...
+        table-replications:
+           - ...
+        transform-options:
+          table-parameters:
+            first-parameter-override: value
+            second-parameter-override: value
+
+This will add `first-table-parameter` and `second-table-parameter` to the metadata of all tables in your replication. These options can be overridden for each replication:
+
+         table-replications:
```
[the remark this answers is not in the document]
max-jacobs: Updated this now, any clearer?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/transformation/TableParametersTransformationTest.java
max-jacobs: Good point, added and all green

CHANGELOG.md
```diff
@@ -1,11 +1,14 @@
...
+## [14.0.2] - TBD
 ### Removed
 * Excluded `org.pentaho:pentaho-aggdesigner-algorithm` from build.
 
 ### Changed
 * Updated `jackson` version to 2.9.9 (was 2.9.8).
 * Updated `circus-train-minimal.yml.template` to include the required `housekeeping` configuration for using the default schema with H2. 
 
+### Added
+* Table transformation to add custom properties to tables during a replication.
```
massdosage: I think this justifies a 14.1.0 release

README.md
```diff
@@ -112,6 +112,13 @@ The preferred way to execute the housekeeping process is by using the `housekeep
 If you want to schedule housekeeping as a separate process then you should use the following script:
 
         $CIRCUS_TRAIN_HOME/bin/housekeeping.sh --config=/path/to/config/file.yml
+        
+### Orphaned data strategy
+To switch off housekeeping completely, and not put any paths to dereferenced data in the housekeeping database, `orphaned-data-strategy` can be set to `NONE`. This property is set to `HOUSEKEEPING` by default. 
+
+Naturally, to substitute housekeeping, Circus Train supports the addition of properties to newly created tables so that you can hook into Hive events to monitor and delete dereferenced data as appropriate. One such tool, [Beekeeper](https://github.com/ExpediaGroup/beekeeper), does exactly this. All that is required if Beekeeper is monitoring your Hive Metastore's events is the addition of Beekeeper specific properties to your configuration (see [Metadata transformations](#metadata-transformations)), and Beekeeper will handle the rest.
```
massdosage: Beekeeper specific -> Beekeeper-specific (?)

README.md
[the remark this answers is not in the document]
massdosage: Yes, ta

README.md
```diff
@@ -565,6 +573,46 @@ The configuration options for Graphite are:
...
+            my-custom-param2: my-custom-value2
+
+In this case, the first table replication will add just one property, `my-custom-override`, while the second will add the two default properties. Adding any number of override properties to a replication will override all default properties. To add properties to your table with non-alphanumeric characters (other than -), surround the key with single quotes and brackets like so:
+
+        transform-options:
+          table-properties:
+            '[my.custom.property2]': my-custom-value
+            '[my.custom.property2]': my-custom-value2
+            
+This feature is a part of Circus Train and no further configuration is required. In addition to this, Circus Train supports custom metadata transformations which are built separately and placed on the Classpath when Circus Train is run. These are documented [below](#metadata-transformations-1).
```
massdosage: Maybe we should rename the section below to "Custom Metadata Transformations" (and update the anchor accordingly) as I think that's the main difference here. In this change you're adding a core on that anyone can use, while the section below describes how users can add their own.

CHANGELOG.md
max-jacobs: Ah yes a minor release, will update

README.md
max-jacobs: Agreed

README.md
max-jacobs: I like it

PR #141, opened 2019-09-04
Improve handling of Avro Table schema replication

### LINKED ISSUE #131: Circus Train should handle Avro Table replication out of the box
Circus Train should be able to detect that a table is an Avro Table with a possibility of an external schema and should trigger the [ct-avro ](https://github.com/HotelsDotCom/circus-train/tree/master/circus-train-avro)transform automatically to copy over the external schema to the replica data lake.

### Context

At the moment, we have a circus-train-avro transform which gets triggered only when the following configuration is provided in the CT config file:

```
transform-options:
    avro-serde-options:
      base-url: s3://shunting-yard-target/bdp/abhi_avro_test
```

If the configuration is not provided, CT treats the replication as a usual replication and as a result the replica table has the parameter `avro.schema.url ` which is pointing to the source table's schema location which is not correct.

### Proposed solution:

CT should be able to detect that the table which is being replicated is a Avro Table and hence should trigger the ct-avro transform and use the table's location as a default location for the schema.

abhimanyugupta07:
Avro table schema will be auto-detected and replicated irrespective of whether avro-serde-options are provided or not. Fixes #131


circus-train-avro/README.md
```diff
@@ -16,6 +16,9 @@ The avro schema file is copied into the following location if the base-url is eq
     
 Note that the `eventId` is generated by circus-train and the `.schema` folder is a hidden folder.
 
+If the specific `avro-serde-options` configuration is not provided by the user, Circus Train will still copy the external schema to the target table in a `.schema` folder.
+
+	${base-url}/eventId/.schema/<avro-file>.avsc
```
max-jacobs: This is the same as above right? Could take line 21 out, and merge line 19 with the above section?

abhimanyugupta07: I have changed it. Does it look better now ?

max-jacobs: Looks great

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 ## TBD
+
+### Added
+* If a user doesn't specify `avro-serde-options`, Circus Train will still copy the external schema over to the target table. Fixes [#131](https://github.com/HotelsDotCom/circus-train/issues/131)
```
AnanaMJ: In other CHANGELOG lines, we have "See #xyz." to refer to the issue number.

```suggestion
* If a user doesn't specify `avro-serde-options`, Circus Train will still copy the external schema over to the target table. See [#131](https://github.com/HotelsDotCom/circus-train/issues/131).
```

PR #145, opened 2019-09-10
Remove pentaho dependency

max-jacobs:
Removing this dependency as I don't think it's available in Maven Central

### DIFF
--- a/circus-train-tool-parent/pom.xml
+++ b/circus-train-tool-parent/pom.xml
@@ -80,12 +80,6 @@
     <dependency>
       <groupId>com.hotels</groupId>
       <artifactId>beeju</artifactId>
-      <exclusions>
-        <exclusion>
-          <artifactId>pentaho-aggdesigner-algorithm</artifactId>
-          <groupId>org.pentaho</groupId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.hive</groupId>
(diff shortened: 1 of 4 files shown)


circus-train-tool-parent/pom.xml
```diff
@@ -80,6 +80,12 @@
     <dependency>
       <groupId>com.hotels</groupId>
       <artifactId>beeju</artifactId>
+      <exclusions>
```
massdosage: Could you rather upgrade to the latest version of BeeJU as this change has been applied there (see BeeJU's CHANGELOG for the 2.0.0 release) so hopefully this dependency no longer comes through transitively?

pom.xml
```diff
@@ -60,7 +60,7 @@
     <junit.version>4.12</junit.version>
     <powermock.version>1.6.4</powermock.version>
     <derby.version>10.10.2.0</derby.version>
-    <beeju.version>1.2.1</beeju.version>
+    <beeju.version>3.0.0</beeju.version>
```
massdosage: 3.0.0 means a move to Java 8 (didn't you see my earlier comment?). Circus Train is still Java 7 (https://github.com/HotelsDotCom/circus-train/blob/master/pom.xml#L47). Can you try 2.0.0 instead?

max-jacobs: As discussed, updated this to instead use Beeju 2.0.0, thanks Adrian

PR #149, opened 2019-10-02
Added TemporaryAWSCredentialsProvider

patduin:
Waiting with merging for some testing.
This hopefully allows users to set per bucket credentials in the core-site.xml and CT will pick up those changes when doing s3MapReduceCopies.


massdosage:
So do I understand correctly that we're probably going to go with https://github.com/HotelsDotCom/circus-train/pull/150 instead of this PR?

patduin:
> So do I understand correctly that we're probably going to go with #150 instead of this PR?

yes I'll close...

PR #150, opened 2019-10-04
add an assume role provider

barnharts4:
@patduin - here are the changes we made.  We branched from master.  Don't need to move to hadoop 2.8.4 for anything.  This was working for all our replications.

We need help with:
1. Reading the role name from `ct.yml` and passing it into the credential provider.
2. Anything we need to do in the pom to either shade or correctly deal with the AWS `sts` jar, and still make it run on-prem correctly.


barnharts4:
@patduin - I took a shot at adding `assume-role` as an option in `copier-options`.  It seemed to fit the pattern of other options perfectly until I needed to read it in our new Provider class.  Then I didn't have a reference to the option name to read it from `Configuration` so I redefined it in the Provider class.  Pretty sure that is not the right way to do it, but I am a bit confused by all the options, parsing, etc. in CT.  Please give me feedback on how to fix.

I did build and run replications to make sure that everything I did works.  Just not sure if it is pretty enough for the PR.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProvider.java
```diff
@@ -0,0 +1,62 @@
...
+public class AssumeRoleCredentialProvider implements AWSCredentialsProvider {
+
+    private static final String ASSUME_ROLE_PROPERTY_NAME = "com.hotels.bdp.circustrain.s3mapreducecp.assumeRole";
+    private static final int CREDENTIALS_DURATION = 12 * 60 * 60; // max duration for assumed role credentials
+
+    private STSAssumeRoleSessionCredentialsProvider.Builder builder;
+    private AWSCredentials credentials;
+
+    public AssumeRoleCredentialProvider(Configuration conf) {
+        checkNotNull(conf, "conf is required");
```
patduin: We should have a try-catch around this. Save the exception and throw it on getCredentials.
If any of these assertions fail during construction `HadoopAWSCredentialProviderChain` can't be created but it should and just fall back to the next in line.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProvider.java
```diff
@@ -0,0 +1,62 @@
...
+        String roleArn = conf.get(ASSUME_ROLE_PROPERTY_NAME);
+        checkArgument(StringUtils.isNotEmpty(roleArn), "Role ARN must not be empty");
```
massdosage: Maybe helpful to put the value of `ASSUME_ROLE_PROPERTY_NAME` in the message so the user knows what property they need to set?

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/HadoopAWSCredentialProviderChain.java
```diff
@@ -42,7 +42,9 @@ public HadoopAWSCredentialProviderChain(String credentialProviderPath) {
   }
 
   public HadoopAWSCredentialProviderChain(Configuration conf) {
-    super(new JceksAWSCredentialProvider(conf), new EC2ContainerCredentialsProviderWrapper());
+    super(new JceksAWSCredentialProvider(conf),
```
massdosage: What's the significance of having the new provider in the `middle` of this chain (as opposed to the end)?

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProviderTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Test(expected = NullPointerException.class)
+    public void throwsNullPointerException() {
+        AssumeRoleCredentialProvider provider = new AssumeRoleCredentialProvider(null);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void throwsIllegalArgumentException() {
+        AssumeRoleCredentialProvider provider = new AssumeRoleCredentialProvider(new Configuration());
+    }
```
massdosage: Can we add a test for the "happy day" scenario when the conf is available and the property is set?

README.md
```diff
@@ -376,6 +376,7 @@ If data is being replicated from HDFS to S3 then Circus Train will use a customi
 | `copier-options.upload-buffer-size`|No|Size of the buffer used to upload the stream of data. If the value is `0` the upload will use the value of the HDFS property `io.file.buffer.size` to configure the buffer. Defaults to `0`|
 | `copier-options.canned-acl`|No|AWS Canned ACL name. See [Access Control List (ACL) Overview](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl) for possible values. If not specified `S3MapReduceCp` will not specify any canned ACL.|
 | `copier-options.copier-factory-class`|No|Controls which copier is used for replication if provided.|
+| `copier-options.assume-role`|No|ARN of an IAM role to assume when writing S3 data to the target replica. Useful when the target is in a different AWS account than CircusTrain is running in. Note that if JCEKS is also configured, JCEKS credentials will be used instead of assuming a role. If `assume-role` is not specified, the copier will use instance credentials.|
```
massdosage: Would it make sense to add this to the s3s3copier too? (can be in a separate PR later).

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/HadoopAWSCredentialProviderChain.java
patduin: it checks them in order, so it will have EC2 container credentials and it will try to use that instead. First one it finds wins

README.md
patduin: yes and yes.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/HadoopAWSCredentialProviderChain.java
massdosage: Yeah, so this changes existing behaviour by putting something higher up in the chain, shouldn't it be added at the end instead?

patduin: no because then always instance profile will be used we'll have no way to override it.
The AssumeRoleCP requires a property to be set otherwise it won't work similar to jceksCP. So I think this is the correct spot. 
Users had jckes configured which will still take priority.
Users relied on instanceProfile won't have the new property so that's ok.
New use case set AssumreRole prop and it will override InstanceProfile.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProvider.java
patduin: I've moved the init code to the refresh serves same purpose but a bit cleaner I think

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProviderTest.java
patduin: Had a quick look so see if we can mock the stsClient but that's not so easy would go down a path of very specific aws object needed to be mocked. If I don't mock we get an error because of course it can't connect. I think we're gonna have to leave that one like it is

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/ConfigurationVariable.java
```diff
@@ -32,6 +34,7 @@ private Constants() {}
 
 public enum ConfigurationVariable {
 
+  ASSUME_ROLE(AssumeRoleCredentialProvider.ASSUME_ROLE_PROPERTY_NAME, null),
```
patduin: I'm using the credential provide property here cause that's what we really need to be setting and we got a dependency on ct-aws not the other way around.

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProviderTest.java
massdosage: OK, yeah, I guess we could come up with some way of passing the client in or having a factory which would by default return the stsClient but for tests we could pass a different one in etc. The code here is very AWS specific and there isn't too much too it so I'm OK with keeping it like this for now.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/HadoopAWSCredentialProviderChain.java
massdosage: OK, that makes sense - do we need to document the order somewhere? Also, would it be possible to have a test that asserts the ordering so someone in future doesn't make a mistake and change the order?

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProviderTest.java
patduin: yeah so that's what I did passing in a mocked Client but then to get then to work I had to mock such specific aws behaviour test would be very brittle

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/HadoopAWSCredentialProviderChain.java
patduin: again test not really possible the list of providers isn't exposed. I'll have a look at the readme see if we can document the order

massdosage: Yeah, OK, I see what you're saying, maybe just a comment here then explaining that order is significant.

README.md
patduin: https://github.com/HotelsDotCom/circus-train/issues/151

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/HadoopAWSCredentialProviderChain.java
barnharts4: I tried last night to create a test for ordering of the chain, but it is a bit difficult since I couldn't figure out a way to mock the `STSAssumeRoleSessionCredentialsProvider` since it is not injected into the wrapper class.  Although maybe something with `spy` would work...

barnharts4: @patduin - thank you for making it lazy fail.  We had it that way originally, I refactored last night when I was trying to write unit tests, then went to bed and immediately realized what I had done.  Bugged me all night and I couldn't sleep. :)

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProviderTest.java
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.aws;
+
+import org.apache.hadoop.conf.Configuration;
+import org.junit.Test;
+
+public class AssumeRoleCredentialProviderTest {
```
abhimanyugupta07: Is it not possible to have a typical() test here ?

patduin: no see the discussion me and @massdosage  had

abhimanyugupta07: Read Adrian's earlier same comment and got my answer.

PR #156, opened 2019-10-18
Fixing transform options and amending README

max-jacobs:
This is a fix to ensure that the default transform options config wires up correctly with the respective class `TransformOptionsConfig` (previously `TableParametersConfg` was used however Spring did not like this).

I've added an integration test with default transforms to make sure.


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptionsConfig.java
```diff
@@ -15,26 +15,15 @@
...
-
-  public static final String TABLE_REPLICATION_TABLE_PARAMETERS = "table-properties";
-
-  private Map<String, String> parameters;
+import com.hotels.bdp.circustrain.api.conf.TransformOptionsConfig;
 
-  public Map<String, String> getParameters() {
-    return parameters;
-  }
+@Configuration("transformOptionsConfig")
```
massdosage: Shouldn't this just be "transformOptions"?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptionsConfig.java
```diff
@@ -15,26 +15,15 @@
...
+@ConfigurationProperties(prefix = TRANSFORM_OPTIONS_PROPERTY)
```
massdosage: Inline the value as called out above.

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/conf/TransformOptionsConfig.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.api.conf;
+
+import java.util.Map;
+
+public class TransformOptionsConfig {
```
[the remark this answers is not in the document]
max-jacobs: Hmm not sure, possibly I was following the pattern set by `AvroSerDeConfig` but that's a bit of a reach. Will update to remove "config", cheers

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptionsConfig.java
```diff
@@ -15,26 +15,15 @@
...
+public class CircusTrainTransformOptionsConfig extends TransformOptionsConfig {
```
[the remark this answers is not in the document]
max-jacobs: Yep agreed

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptionsConfig.java
max-jacobs: This value maps the configuration to a bean, so it should really match the bean the config is being mapped to. Now that the class has changed name, it should indeed be just `transformOptions`.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptionsConfig.java
max-jacobs: not sure I follow here, do you mean `@ConfigurationProperties(prefix = TRANSFORM_OPTIONS_PROPERTY)` -> `@ConfigurationProperties(TRANSFORM_OPTIONS_PROPERTY)`

or are you referring to `@Configuration("transformOptions")` -> `@Configuration(TRANSFORM_OPTIONS)`

thanks

max-jacobs: ah I think I follow you, moving the properties from api to core so that they can be referenced here?

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/conf/TransformOptionsConfig.java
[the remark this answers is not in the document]
max-jacobs: done

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/conf/TransformOptionsConfig.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public static final String TRANSFORM_OPTIONS_PROPERTY = "transform-options";
```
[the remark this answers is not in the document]
max-jacobs: done

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptionsConfig.java
max-jacobs: and I think done

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptions.java
```diff
@@ -15,26 +15,20 @@
...
+@Configuration(TRANSFORM_OPTIONS_BEAN)
+@ConfigurationProperties(TRANSFORM_OPTIONS_PROPERTY)
+public class CircusTrainTransformOptions extends TransformOptions {
 
-  public Map<String, String> getParameters() {
-    return parameters;
-  }
+  public static final String TRANSFORM_OPTIONS_BEAN = "transformOptions";
+  public static final String TRANSFORM_OPTIONS_PROPERTY = "transform-options";
+  public static final String TABLE_REPLICATION_TABLE_PARAMETERS = "table-properties";
```
massdosage: Should we not just call this `TABLE_PROPERTIES` here and only refer to them as "params" when they get closer to Hive? I remember us agreeing on some convention here but can't remember exactly where the boundary was in property vs param.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptions.java
```diff
@@ -15,26 +15,20 @@
...
-@Configuration
-@ConfigurationProperties(prefix = TABLE_REPLICATION_TABLE_PARAMETERS)
-public class TableParametersConfig {
-
-  public static final String TABLE_REPLICATION_TABLE_PARAMETERS = "table-properties";
+import com.hotels.bdp.circustrain.api.conf.TransformOptions;
 
-  private Map<String, String> parameters;
+@Configuration(TRANSFORM_OPTIONS_BEAN)
+@ConfigurationProperties(TRANSFORM_OPTIONS_PROPERTY)
```
massdosage: Why does this no longer need the `prefix = ` bit? The build passes in Travis so it doesn't appear to matter, I just can't remember exactly how this works...

max-jacobs: I think the default argument is `value` and they're interchangeable. So you can either just put a string straight in and Spring will set prefix to `value`, or you can specify `prefix = ...`

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/conf/CircusTrainTransformOptions.java
max-jacobs: I think everything everything external is a property, and internal is a param, and the boundary is as close to the user as possible. So yep I'll do `TABLE_REPLICATION_TABLE_PARAMETERS` -> `TABLE_PROPERTIES`, and then everything which uses this will be called a param.

massdosage: :+1:

max-jacobs:
@massdosage, @patduin has made a good point that the `avro-serde-options` transform, when used in a table replication has this format:

```

**Comment by max-jacobs** [comment]
@massdosage, @patduin has made a good point that the avro-serde-options transform, when used in a table replication has this format:

```
table-replications:
  - source-table:
      blah
    replica-table:
      blah
    transform-options:
      avro-serde-options:
        blah
```

but when a default is provided, it takes this format:

```
table-replications:
  - source-table:
      blah
    replica-table:
      blah
avro-serde-options:
  blah
```

i.e. the "transform-options" qualifier is no longer needed. This PR is based on the assumption that the "transform-options" qualifier should be present.

So we can either merge this and update the avro options to follow the same pattern (and make a breaking change), or we can revert some of the changes here so that we stay in line with how the avro options work.

I'll update the readme in either case and keep the integration test so that it's clear what's going on. Thanks!

File: CHANGELOG.md [hunk 8]
```diff
@@ -1,5 +1,10 @@
-## TBD
-* Updated `hotels-oss-parent` version to 4.2.0 (was 4.0.0). Contains updates to the copyright header.
```
patduin: @massdosage, @patduin has made a good point that the avro-serde-options transform, when used in a table replication, has this format:

table-replications:
  - source-table:
      blah
    replica-table:
      blah
    transform-options:
      avro-serde-options:
        blah
but when a default is provided, it takes this format:

table-replications:
  - source-table:
      blah
    replica-table:
      blah
avro-serde-options:
  blah
i.e. the "transform-options" qualifier is no longer needed. This PR is based on the assumption that the "transform-options" qualifier should be present.

So we can either merge this and update the avro options to follow the same pattern (and make a breaking change), or we can revert some of the changes here so that we stay in line with how the avro options work.

I'll update the readme in either case and keep the integration test so that it's clear what's going on. Thanks!

patduin: Putting your comment here so we can more easily reply on it

patduin: I'm in favour of keeping it the same as the avro-serde-config. I'm not sure adding transform-options adds that much. I think this PR as it is now is already a breaking change right as you've changed some of the public API classes.

abhimanyugupta07: +1 for keeping it same as `avro-serde-config` coz as a user I would expect if I have provided a global setting, it would take that as default unless another config is specified for that particular `replica-table`.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+## [15.0.0] - TBD
+
+### Updated
```
abhimanyugupta07: `Updated` is not a valid tag as per https://keepachangelog.com/en/1.0.0/.

This should go in `Changed`.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/partitioned-single-table-avro-schema-metadata-update.yml
```diff
@@ -7,5 +7,6 @@ table-replications:
       partition-limit: 100
     replica-table:
       table-name: ct_table_p_managed_copy
-avro-serde-options:
-  base-url: ${circus-train-runner.replica-warehouse-uri}${circus-train-runner.database-name}/
\ No newline at end of file
+transform-options:
+  avro-serde-options:
+    base-url: ${circus-train-runner.replica-warehouse-uri}${circus-train-runner.database-name}/
```
abhimanyugupta07: missing newlines in some of the YML files

max-jacobs: Yeah that's how I found them, but I'll add them in

patduin:
code looks ok guess it is just deciding if we want to do this breaking change now and update version/README etc...

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
-## TBD
-* Updated `hotels-oss-parent` version to 4.2.0 (was 4.0.0). Contains updates to the copyright header.
+## [15.0.0] - TBD
+
+### Fixed
+* Table properties can now be added to default transformations.
+
+### Changed
+* `hotels-oss-parent` version to 4.2.0 (was 4.0.0). Contains updates to the copyright header.
+* Default `avro-serde-options` must now be included within `transform-options`.
```
massdosage: This is the biggest change impacting downstream users so put the Changed block first and then this line first within the block.

massdosage: Add a note that this is backwards incompatible config change.

massdosage: Link to avro README file for the example.

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformation.java
```diff
@@ -15,27 +15,37 @@
...
   private Map<String, Object> avroSerdeConfigOverride = Collections.emptyMap();
   static final String AVRO_SCHEMA_URL_PARAMETER = "avro.schema.url";
 
-  protected AbstractAvroSerDeTransformation(AvroSerDeConfig avroSerDeConfig) {
-    this.avroSerDeConfig = avroSerDeConfig;
+  protected AbstractAvroSerDeTransformation(TransformOptions transformOptions) {
+    if (transformOptions.getTransformOptions() == null) {
+      return;
+    }
+    Object avroSerDeOptionsObj = transformOptions.getTransformOptions().get(AVRO_SERDE_OPTIONS);
```
massdosage: Add a test for this being null.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/unpartitioned-single-table-avro-schema.yml
```diff
@@ -6,5 +6,9 @@ table-replications:
     replica-table:
       table-name: ct_table_u_managed_copy
       table-location: ${circus-train-runner.replica-warehouse-uri}${circus-train-runner.database-name}/ct_table_u_managed_copy
-avro-serde-options:
-  base-url: ${circus-train-runner.replica-warehouse-uri}${circus-train-runner.database-name}/
\ No newline at end of file
+transform-options:
```
massdosage: Instead of re-using this file, just create a new one which has your multiple transformations in it.

CHANGELOG.md
max-jacobs: Done

circus-train-avro/src/main/java/com/hotels/bdp/circustrain/avro/transformation/AbstractAvroSerDeTransformation.java
max-jacobs: Done, for both transformations

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/unpartitioned-single-table-avro-schema.yml
max-jacobs: Separated, and added a new IT for transformation overrides

max-jacobs:
I think this is ready now, just running all the tests locally

max-jacobs:
This might actually need one more change. Spring seems to be converting `true` in the config to a Boolean, which throws a `ClassCastException` when the replica table is created. This can be avoided by putting `"true"` in the config instead to force it in to a String, or I can add a check to cast properly to a String when the transformation is applied. What do people think?

massdosage:
> This might actually need one more change. Spring seems to be converting `true` in the config to a Boolean, which throws a `ClassCastException` when the replica table is created. This can be avoided by putting `"true"` in the config instead to force it in to a String, or I can add a check to cast properly to a String when the transformation is applied. What do people think?

Where in the config is this boolean?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
-## TBD
-* Updated `hotels-oss-parent` version to 4.2.0 (was 4.0.0). Contains updates to the copyright header.
+## [15.0.0] - TBD
+
+### Changed
```
max-jacobs: @massdosage adding a comment here to discuss
>>This might actually need one more change. Spring seems to be converting true in the config to a Boolean, which throws a ClassCastException when the replica table is created. This can be avoided by putting "true" in the config instead to force it in to a String, or I can add a check to cast properly to a String when the transformation is applied. What do people think?

>Where in the config is this boolean?

For example:
```
transform-options:
  table-properties:
    '[some.property]': true
```
Spring will read that property in to a Boolean, and then even though the map is cast to `Map<String, String>`, it remains a Boolean.  It's only when the table is created in Hive that the `ClassCastException` happens.

So we could validate before that happens, or just tell the user to wrap their values in quotes

max-jacobs: I kind of now recall us discussing this at some point (?) and I think we thought this is up to the user to enter their config correctly and not for Circus Train to decide.

massdosage: As much as I like to have proper types wherever possible I think we don't really want to have to consider all the types that could be passed into these maps so we could just document that the keys and values need to be **Strings** and then the end users can do conversion from Strings into whatever types they need themselves in their transforms.

max-jacobs: Okay cool, I've added to the readme. Thanks!

README.md
```diff
@@ -609,10 +609,10 @@ This will add `my-custom-property` and `my-custom-property2` to the metadata of
                ...
          transform-options:
            table-properties:
-            my-custom-param: my-custom-value
-            my-custom-param2: my-custom-value2
+            my-custom-property: my-custom-value
```
patduin: if I specify 1 properties in my table replication and I also specify 1 different property in the "global" one. I think from looking at the code the 1 table replication property is then ignored. Is this what we want?

max-jacobs: From the readme, any properties in a table replication will override the global properties. Where in the code do you see that it is ignored? 

On [line 44 in AbstractTableParametersTransformation](https://github.com/HotelsDotCom/circus-train/pull/156/files#diff-b75474f19d562493acfb0d036ff0178cR44) the override is returned if they exist, instead of the global.

patduin: that's my point it's either all global ones or all local ones. I can't set 2 global properties and override just 1 in a replication. Wouldn't it make more sense to take the whole local map and add all override key/values??

abhimanyugupta07: Hmm.. thinking from a user perspective, how will he set certain properties for his one particular table when he doesn't want to add global properties ?

I incline towards letting user copy some of the global properties to table level if he needs both, at least then we support both use-cases.

patduin: ok I agree, in the alternative you'd have to override and set to null or something. All good.

circus-train-avro/README.md
```diff
@@ -42,10 +42,11 @@ the global level, then be overriden for certain replications.
...
           table-location: /example/url
-        
-    avro-serde-options:
-      base-url: /example/url/
-      #Schema url's for each table will be copied to /example/url/<eventId>/.schema/
+    
+    transform-options:    
+      avro-serde-options:
+        base-url: /example/url/
+        #Schema url's for each table will be copied to /example/url/<eventId>/.schema/
```
massdosage: NABD but `url's` should be `URLs`

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/unpartitioned-single-table-multiple-transformation-overrides.yml
```diff
@@ -0,0 +1,20 @@
...
+        base-url: ${circus-train-runner.replica-warehouse-uri}${circus-train-runner.database-name}/override
+      table-properties:
+        '[table.property.first]': first-override
+        '[table.property.second]': second-override
+transform-options:
+  avro-serde-options:
+    base-url: ${circus-train-runner.replica-warehouse-uri}${circus-train-runner.database-name}/
+  table-properties:
+    '[table.property.first]': first
+    '[table.property.second]': second
```
massdosage: Closing line break plz

Issue #151, opened 2019-10-04
Provide sts assume role support for s3-s3 copier


patduin:
_As a user of CT_
_I'd like to configure ct with a user arn for source and target_
_So that CT can assume the correct role when reading and writing data_

Similar to what we provide to s3MapReduceCopier we should allow the sts assume credentials provider to be configured. 
It should be configurable on source and target with different settings.

massdosage:
Code changes will probably be very similar to the other half of https://github.com/HotelsDotCom/circus-train/pull/105/files <- this changed both the `S3MapReduceCopier` (which has already been done in the PR referenced above) and the `S3S3Copier` (what needs to happen here).

PR #157, opened 2019-10-25
S3s3 assume role provider

JayGreeeen:
Adds functionality for assuming roles when replicating between s3.


circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProvider.java
```diff
@@ -29,7 +29,7 @@
 public class AssumeRoleCredentialProvider implements AWSCredentialsProvider {
 
   public static final String ASSUME_ROLE_PROPERTY_NAME = "com.hotels.bdp.circustrain.aws.AssumeRoleCredentialProvider.assumeRole";
-  private static final int CREDENTIALS_DURATION = 12 * 60 * 60; // max duration for assumed role credentials
+  private static final int CREDENTIALS_DURATION = 60 * 60; // max duration for assumed role credentials
```
massdosage: What unit is this measured in? Perhaps add this to the variable name or at the very least a comment stating this. I know this code predates your change but we might as well fix it.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
```diff
@@ -292,7 +293,7 @@ private Copy submitCopyJob(CopyJobRequest copyJob) {
                   copyObjectRequest.getSourceBucketName(),
                   copyObjectRequest.getSourceKey());
           LOG
-              .debug("Copy failed with exception:", e);
+              .info("Copy failed with exception:", e);
```
massdosage: I'm uncertain why this was on debug, maybe because if it fails it tries again? It feels more like it should be on warn but I don't know how often this happens and how noisy it could be? Alternatively put it back the way it was.

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierOptionsTest.java
```diff
@@ -115,5 +115,19 @@ public void getMaxCopyAttemptsDefaultIfLessThanOne() throws Exception {
...
+    copierOptions.put(S3S3CopierOptions.Keys.ASSUME_ROLE.keyName(), "iam:role:1234:user");
+    S3S3CopierOptions options = new S3S3CopierOptions(copierOptions);
+    assertThat(options.getAssumedRole(), is("iam:role:1234:user"));
+  }
+  
+  @Test
+  public void getAssumedRoleDefaultIsNull() throws Exception {
+    S3S3CopierOptions options = new S3S3CopierOptions(copierOptions);
+    assertNull(options.getAssumedRole());
+
```
massdosage: remove blank line

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/S3S3CopierTest.java
```diff
@@ -74,6 +75,7 @@
 import com.hotels.bdp.circustrain.s3s3copier.aws.ListObjectsRequestFactory;
 import com.hotels.bdp.circustrain.s3s3copier.aws.TransferManagerFactory;
 
+@Ignore
```
massdosage: WHAT?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
```diff
@@ -58,6 +67,25 @@ public AmazonS3 newInstance(AmazonS3URI uri, S3S3CopierOptions s3s3CopierOptions
     }
   }
 
+  public AmazonS3 newTargetInstance(AmazonS3URI uri, S3S3CopierOptions s3s3CopierOptions) {
```
massdosage: Does this need to be public?

massdosage: It's not in the interface so that's why you've had to cast the factory to the concrete type when you use it in the actual s3s3 copier. Do you really have to do that or can `newInstance` do what is necessary?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
```diff
@@ -38,10 +40,17 @@
   private final static Logger LOG = LoggerFactory.getLogger(JceksAmazonS3ClientFactory.class);
 
   private final Security security;
+  private final HiveConf conf;
 
   @Autowired
+  public JceksAmazonS3ClientFactory(Security security, HiveConf replicaHiveConf) {
```
massdosage: Why do we need the hive configuration passed in now?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
```diff
@@ -38,10 +40,17 @@
...
+    this.security = security;
+    this.conf = replicaHiveConf;
+  }
+
   public JceksAmazonS3ClientFactory(Security security) {
     this.security = security;
+    this.conf = new HiveConf();
```
massdosage: There can be side effects of calling new on a HiveConf so it would be probably be better if we just set/left it null if not set (if we really have to have it).

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
```diff
@@ -58,6 +67,25 @@ public AmazonS3 newInstance(AmazonS3URI uri, S3S3CopierOptions s3s3CopierOptions
     }
   }
 
+  public AmazonS3 newTargetInstance(AmazonS3URI uri, S3S3CopierOptions s3s3CopierOptions) {
+    AmazonS3 globalClient;
+
+    String assumedRole = s3s3CopierOptions.getAssumedRole();
+    if (assumedRole != null) {
```
massdosage: There's a lot of code in this method which is duplicated elsewhere in this class, feels like a lot of overlap between methods. To me it feels like there should be two code paths - one where the role is assumed (the new stuff) and one where it isn't (what was there before). Can we try isolate the differences into a few new methods based on an if statement on `assumedRole` being null and everything else can stay as it was.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProvider.java
massdosage: Why are we even changing this?

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/S3S3Copier.java
JayGreeeen: It tries 3 times by default so could end up printing quite a lot out

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
JayGreeeen: The role is only used for the target client and theres no way to differentiate between the source and target, so there needs to be separate methods

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
JayGreeeen: The role is added to the configuration, and thats then given to the credential provider chain. It's needed so that `AssumeRoleCredentialProvider` can be used

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AssumeRoleCredentialProvider.java
JayGreeeen: Its in seconds, I've added it to the comment. Think Abhi changed this when it was timing out and we were trying to figure out why. Will change it back

JayGreeeen: Found out why we changed it: the role we're using for testing has a max duration of 1 hour, but this duration was set to 12 hours so it was failing. Will need to increase the duration of the role.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
patduin: We can either add this method to the interface or instead we can create two beans: TargetClientFactory and SourceClientFactory which can be implemented by this class but with different configuration. For instance this takes the ReplicaHiveConf so this would be the TargetClientFactory. Happy to discuss :)

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/AmazonS3ClientFactory.java
```diff
@@ -28,6 +27,7 @@
    * @param s3s3CopierOptions Copier options.
    * @return a new instance of {@linkplain AmazonS3 AmazonS3}
    */
-  AmazonS3 newInstance(AmazonS3URI uri, S3S3CopierOptions s3s3CopierOptions);
+  AmazonS3 newSourceInstance(AmazonS3URI uri, S3S3CopierOptions s3s3CopierOptions);
 
+  AmazonS3 newTargetInstance(AmazonS3URI uri, S3S3CopierOptions s3s3CopierOptions);
```
massdosage: Add javadoc here too plus a one liner description above the params that outlines what they do (mainly the difference between "source" and "target" factories).

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
abhimanyugupta07: Do we need source and target client separate still ?

If a `assume-role` is provided, we could just use a client with that assumed role coz it will have read access on the source. So, I feel like, this class could just have `newInstance` method and based on `s3s3copierOptions`, different type of client can be returned to be used either as source or target.

I will sit with Jay and think more about it tomorrow morning.

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
abhimanyugupta07: I was trying out few things to refactor this and ended up creating this PR: https://github.com/HotelsDotCom/circus-train/pull/159

Basically, the switch now happens only at `getCredentialsProviderChain` function. Have a look and we can improve it further.

README.md
```diff
@@ -390,6 +390,7 @@ If data is being replicated from S3 to S3 then Circus Train will use the AWS S3
 |`copier-options.canned-acl`|No|AWS Canned ACL name. See [Access Control List (ACL) Overview](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl) for possible values. If not specified `S3S3Copier` will not specify any canned ACL.|
 |`copier-options.copier-factory-class`|No|Controls which copier is used for replication if provided.|
 |`copier-options.s3s3-retry-max-copy-attempts`|No|Controls the maximum number of attempts if AWS throws an error during copy. Default value is 3.|
+| `copier-options.assume-role`|No|ARN of an IAM role to assume when writing S3 data to the target S3 location. Useful when the target is in a different AWS account than CircusTrain is running in. Note that if JCEKS is also configured, JCEKS credentials will be used instead of assuming a role. If `assume-role` is not specified, the copier will use instance credentials. The role provided must have read access to the S3 source and write access to the S3 target.|
```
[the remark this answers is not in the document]
JayGreeeen: It is mentioned at the end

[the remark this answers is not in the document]
abhimanyugupta07: Oh. I missed that. 👍

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## TBD
 * Updated `hotels-oss-parent` version to 4.2.0 (was 4.0.0). Contains updates to the copyright header.
```
max-jacobs: Not sure why this doesn't have a heading already, but this should be under a `### Changed`

README.md
max-jacobs: CircusTrain -> Circus Train

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
```diff
@@ -38,20 +41,43 @@
...
+      HadoopAWSCredentialProviderChain credentialProviderChain) {
     LOG.debug("trying to get a client for uri '{}'", uri);
-    AmazonS3 globalClient = newGlobalInstance(s3s3CopierOptions);
+    AmazonS3 globalClient = newGlobalInstance(s3s3CopierOptions, credentialProviderChain);
     try {
+
+      /*
+       * When using roles it can take a while for the credentials to be retrieved from the
+       * AssumeRoleCredentialsProvider. This can mean that the rest of the code completes before the credentials are
+       * retrieved, resulting in errors. A temporary fix for this situation is to put the thread to sleep for 10s to
```
max-jacobs: Where is this sleep happening? Is it not part of this PR, just a recommendation?

README.md
max-jacobs: Just seen that its written the same for the other assume role config, perhaps we should change both? Or just leave it. I don't mind.

JayGreeeen: I'll change any occurrences of 'CircusTrain' to have a space so its consistent with the rest of the readme

circus-train-s3-s3-copier/src/main/java/com/hotels/bdp/circustrain/s3s3copier/aws/JceksAmazonS3ClientFactory.java
JayGreeeen: It's not currently happening - we added it in to fix the issue mentioned but then changed the code and found it was no longer needed. It was suggested that we keep it in a comment just in case any one comes across the issue again. It's a bit of a weird one though, and the comment seems to be confusing a few people so unsure if its all that helpful.

JayGreeeen:
fixes #151

PR #165, opened 2020-01-24
applying limit when creating the partition filter

### LINKED ISSUE #164: partitionLimit does not trim partitionPredicate
When running with a generated partition (https://github.com/HotelsDotCom/circus-train#partition-filter-generation) the partitionLimit flag is not used when generating the partitionFilter. As a consequence a table with a large number of partitions might generate a huge filter string that is sent to the metastore. 
The metastore call (org.apache.hadoop.hive.metastore.HiveMetaStoreClient.listPartitionsByFilter) does have apply the limit but will fail when trying to parse the filter.
It would be better if we'd prevent such a huge filter string and apply such a limit much earlier on in the code path.

patduin:
fixed #164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
 ### Fixed
-Fixed issue were partition-limit is not correctly applied when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
+Fixed issue where partition-limit is not correctly applied when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
 
 ## [15.0.0] - 2019-11-12
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## [15.1.0] - TBD
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
-* Changed the order of the generated partition filter (HiveDiff) it is now reverse natural order (which means new partitions first when partitions are date/time strings). When in doubt use the circus-train-tool `check-filters.sh` to so what would be generated.
+* Changed the order of the generated partition filter used by "HiveDiff"  - it is now reverse natural order (which means new partitions first when partitions are date/time strings). When in doubt use the circus-train-tool `check-filters.sh` to see what would be generated.
 
 ### Fixed
 Fixed issue where partition-limit is not correctly applied when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [15.1.0] - TBD
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
+
 ### Fixed
 Fixed issue where partition-limit is not correctly applied when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
 ### Fixed
-Fixed issue were limit is ignore when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
+Fixed issue were partition-limit is not correctly applied when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
 
 ## [15.0.0] - 2019-11-12
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## [15.1.0] - TBD
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
+### Fixed
+Fixed issue were limit is ignore when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
 
 ## [15.0.0] - 2019-11-12
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [15.1.0] - TBD
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
+* Changed the order of the generated partition filter (HiveDiff) it is now reverse natural order (which means new partitions first when partitions are date/time strings). When in doubt use the circus-train-tool `check-filters.sh` to so what would be generated.
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## [15.1.0] - TBD
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
+### Fixed
+Fixed issue were limit is ignore when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
```
max-jacobs:
```suggestion

Fixed issue where limit is ignore when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/DiffGeneratedPartitionPredicate.java
```diff
@@ -45,6 +46,7 @@
   private final Function<Path, String> checksumFunction;
   private String partitionPredicate;
   private boolean generated = false;
+  private Short partitionLimit = -1;
```
max-jacobs: How comes this isn't initialised to `UNLIMITED` here? This is only to figure out the limit on line 183 in `HiveDifferences` right?

CHANGELOG.md
patduin: I've reworded it a bit

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## [15.1.0] - TBD
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
+### Fixed
+Fixed issue were partition-limit is not correctly applied when generating a partition filter. See [#164](https://github.com/HotelsDotCom/circus-train/issues/164).
```
nvitucci: Typo: "were" -> "where"

circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/HiveDifferences.java
```diff
@@ -162,14 +171,16 @@ private HiveDifferences(
...
+      Function<Path, String> checksumFunction,
+      int partitionLimit) {
     this.diffListener = diffListener;
     this.comparatorRegistry = comparatorRegistry;
     this.sourceTable = sourceTable;
     this.sourcePartitionIterator = sourcePartitionIterator;
     this.replicaTable = replicaTable;
     this.replicaPartitionFetcher = replicaPartitionFetcher;
     this.checksumFunction = checksumFunction;
+    this.partitionLimit = partitionLimit < 0 ? UNLIMITED : partitionLimit;
```
nvitucci: What if the `partitionLimit` is 0?

patduin: then it's 0, and stays 0 and you get 0 partitions back just like you asked :)

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/DiffGeneratedPartitionPredicate.java
patduin: UNLIMITED is max int
partitionLimit for hive fetch partition call is -1 for being unlimited.
So I map -1 to max int in HiveDiferences so my loop nicely reads while (i < limit) https://github.com/HotelsDotCom/circus-train/blob/3a5e3bfaefe28ddcd500fa8a3451f6489005b2e7/circus-train-comparator/src/main/java/com/hotels/bdp/circustrain/comparator/hive/HiveDifferences.java#L219
 it's mostly sugarcoating to be honest...

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## [15.1.0] - TBD
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
+### Fixed
```
max-jacobs: Think you also need a line between Changed and Fixed?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/DiffGeneratedPartitionPredicate.java
```diff
@@ -55,19 +57,23 @@ public DiffGeneratedPartitionPredicate(
...
+    }
   }
 
   private String generate() {
     try (CloseableMetaStoreClient sourceMetastore = source.getMetaStoreClientSupplier().get()) {
       try (CloseableMetaStoreClient replicaMetastore = replica.getMetaStoreClientSupplier().get()) {
         Table sourceTable = source.getTableAndStatistics(tableReplication).getTable();
         PartitionIterator partitionIterator = new PartitionIterator(sourceMetastore, sourceTable,
-            tableReplication.getPartitionIteratorBatchSize());
+            tableReplication.getPartitionIteratorBatchSize(), Ordering.REVERSE);
```
massdosage: Isn't this also a change in behaviour that we should mention in the CHANGELOG? HiveDiff will now process the partitions with newest first while in the past it was oldest first?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/DiffGeneratedPartitionPredicateTest.java
```diff
@@ -106,17 +105,23 @@ public void setUp() throws Exception {
   public void autogeneratePredicate() throws Exception {
     when(replica.getTableAndStatistics(tableReplication)).thenReturn(replicaTableAndStats);
     when(replicaTableAndStats.getTable()).thenReturn(table2);
+    when(sourceTable.getPartitionLimit()).thenReturn((short) 10);
+
+    predicate = new DiffGeneratedPartitionPredicate(source, replica, tableReplication, checksumFunction);
 
     assertThat(predicate.getPartitionPredicate(),
-        is("(p1='value1' AND p2='value2') OR (p1='value11' AND p2='value22')"));
+        is("(p1='value11' AND p2='value22') OR (p1='value1' AND p2='value2')"));
```
massdosage: I can't make up my mind whether this would be considered a breaking change. Technically I think it is, but I guess most of our end users wouldn't notice?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/DiffGeneratedPartitionPredicate.java
patduin: yeah I'll add a note

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/DiffGeneratedPartitionPredicateTest.java
patduin: I was doubting if we should make it configurable, but to be honest i you're using hiveDiff and your whole table is copied its all the same. It's just for the backfill, I think it's not a breaking change.

massdosage: OK, then let's just document it in the CHANGELOG. If there is an easy place to mention the order that HiveDiff processes the partitions in the README that would be nice but not required. We could add a feature to make it configurable in the future if anyone ever actually needs it.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [15.1.0] - TBD
 ### Changed
 * AVRO Schema Copier now re-uses the normal 'data' copier instead of its own. See [#162](https://github.com/HotelsDotCom/circus-train/issues/162). 
+* Changed the order of the generated partition filter (HiveDiff) it is now reverse natural order (which means new partitions first when partitions are date/time strings). When in doubt use the circus-train-tool `check-filters.sh` to so what would be generated.
```
massdosage:
```suggestion
* Changed the order of the generated partition filter used by "HiveDiff"  - it is now reverse natural order (which means new partitions first when partitions are date/time strings). When in doubt use the circus-train-tool `check-filters.sh` to see what would be generated.
```

Issue #166, opened 2020-02-05
If table has large number of partitions to replicate, Replica.updateMetadata should send partition list in batches


barnharts4:
**Is your feature request related to a problem? Please describe.**
When replicating a table with over 40,000 partitions, replication fails in `Replica.updateMetadata` with a timeout on the Hive metastore, due to sending a `List<Partition>` of size 40000 in one thrift request.

**Describe the solution you'd like**
The list of partitions to be added/altered should be broken up into manageable batch sizes so as not to overwhelm the hive metastore with a huge request.

patduin:
Thanks for contributing!

PR #167, opened 2020-02-05
Issue #166 - Replica.updateMetastore calls metastore with batched partition lists.

barnharts4:
Fixes Issue #166 

Log file output from running test on table with 5019 partitions:
```
21:01:53.209 [main] INFO  com.hotels.bdp.circustrain.core.replica.Replica - Updating replica table metadata.
21:01:53.334 [main] INFO  com.hotels.hcommon.hive.metastore.client.closeable.CloseableMetaStoreClientInvocationHandler - Couldn't invoke method public abstract org.apache.hadoop.hive.metastore.api.Table org.apache.hadoop.hive.metastore.IMetaStoreClient.getTable(java.lang.String,java.lang.String) throws org.apache.hadoop.hive.metastore.api.MetaException,org.apache.thrift.TException,org.apache.hadoop.hive.metastore.api.NoSuchObjectException
21:11:02.672 [main] INFO  com.hotels.bdp.circustrain.core.replica.Replica - Creating 5019 new partitions.
21:11:02.677 [main] INFO  com.hotels.bdp.circustrain.core.replica.Replica - Creating partitions 0 through 999
21:15:21.485 [main] INFO  com.hotels.bdp.circustrain.core.replica.Replica - Creating partitions 1000 through 1999
21:19:35.692 [main] INFO  com.hotels.bdp.circustrain.core.replica.Replica - Creating partitions 2000 through 2999
21:23:57.927 [main] INFO  com.hotels.bdp.circustrain.core.replica.Replica - Creating partitions 3000 through 3999
21:28:13.187 [main] INFO  com.hotels.bdp.circustrain.core.replica.Replica - Creating partitions 4000 through 4999
21:32:30.142 [main] INFO  com.hotels.bdp.circustrain.core.replica.Replica - Creating partitions 5000 through 5018
21:32:35.818 [main] INFO  hive.metastore - Closed a connection to metastore, current connections: 0
21:32:35.818 [main] INFO  com.hotels.bdp.circustrain.core.PartitionedTableReplication - Replicated 5019 partitions of table test.barnparttest5000.
```

### DIFF
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
@@ -30,8 +30,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Lists;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.conf.HiveConf;
@@ -44,9 +42,11 @@ import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Enums;
 import com.google.common.base.Optional;
 import com.google.common.base.Supplier;
+import com.google.common.collect.Lists;
 
 import com.hotels.bdp.circustrain.api.CircusTrainException;
 import com.hotels.bdp.circustrain.api.ReplicaLocationManager;
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
@@ -67,14 +67,13 @@ import com.hotels.hcommon.hive.metastore.util.LocationUtils;
 public class Replica extends HiveEndpoint {
 
   private static final Logger LOG = LoggerFactory.getLogger(Replica.class);
-  private static final int DEFAULT_PARTITION_BATCH_SIZE = 1000;
 
   private final ReplicaTableFactory tableFactory;
   private final HousekeepingListener housekeepingListener;
   private final ReplicaCatalogListener replicaCatalogListener;
   private final ReplicationMode replicationMode;
   private final TableReplication tableReplication;
-  private final int partitionBatchSize;
+  private int partitionBatchSize = 1000;
 
   /**
    * Use {@link ReplicaFactory}
@@ -93,7 +92,6 @@ public class Replica extends HiveEndpoint {
     this.housekeepingListener = housekeepingListener;
     replicationMode = tableReplication.getReplicationMode();
     this.tableReplication = tableReplication;
-    this.partitionBatchSize = DEFAULT_PARTITION_BATCH_SIZE;
   }
 
   /**
@@ -226,7 +224,7 @@ public class Replica extends HiveEndpoint {
           for (List<Partition> sublist : Lists.partition(partitionsToAlter, partitionBatchSize)) {
             int start = counter * partitionBatchSize;
             LOG.info("Altering partitions {} through {}", start, start + sublist.size() - 1);
-            client.alter_partitions(replicaDatabaseName, replicaTableName, partitionsToAlter);
+            client.alter_partitions(replicaDatabaseName, replicaTableName, sublist);
             counter++;
           }
         } catch (TException e) {
@@ -246,7 +244,7 @@ public class Replica extends HiveEndpoint {
           for (List<ColumnStatistics> sublist : Lists.partition(statisticsToSet, partitionBatchSize)) {
             int start = counter * partitionBatchSize;
             LOG.info("Setting column statistics for partitions {} through {}", start, start + sublist.size() - 1);
-            client.setPartitionColumnStatistics(new SetPartitionsStatsRequest(statisticsToSet));
+            client.setPartitionColumnStatistics(new SetPartitionsStatsRequest(sublist));
             counter++;
           }
         } catch (TException e) {
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2019 Expedia, Inc.
+ * Copyright (C) 2016-2020 Expedia, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,6 +30,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.conf.HiveConf;
@@ -65,12 +67,14 @@ import com.hotels.hcommon.hive.metastore.util.LocationUtils;
 public class Replica extends HiveEndpoint {
 
   private static final Logger LOG = LoggerFactory.getLogger(Replica.class);
+  private static final int DEFAULT_PARTITION_BATCH_SIZE = 1000;
 
   private final ReplicaTableFactory tableFactory;
   private final HousekeepingListener housekeepingListener;
   private final ReplicaCatalogListener replicaCatalogListener;
   private final ReplicationMode replicationMode;
   private final TableReplication tableReplication;
+  private final int partitionBatchSize;
 
   /**
    * Use {@link ReplicaFactory}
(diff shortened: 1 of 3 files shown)


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -189,7 +222,13 @@ public void updateMetadata(
       if (!partitionsToAlter.isEmpty()) {
         LOG.info("Altering {} existing partitions.", partitionsToAlter.size());
         try {
-          client.alter_partitions(replicaDatabaseName, replicaTableName, partitionsToAlter);
+          int counter = 0;
+          for (List<Partition> sublist : Lists.partition(partitionsToAlter, partitionBatchSize)) {
+            int start = counter * partitionBatchSize;
+            LOG.info("Altering partitions {} through {}", start, start + sublist.size() - 1);
+            client.alter_partitions(replicaDatabaseName, replicaTableName, partitionsToAlter);
```
abhimanyugupta07: shouldn't this be `sublist` instead of `partitionsToAlter`.

`client.alter_partitions(replicaDatabaseName, replicaTableName, partitionsToAlter);`

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -203,7 +242,13 @@ public void updateMetadata(
       if (!statisticsToSet.isEmpty()) {
         LOG.info("Setting column statistics for {} partitions.", statisticsToSet.size());
         try {
-          client.setPartitionColumnStatistics(new SetPartitionsStatsRequest(statisticsToSet));
+          int counter = 0;
+          for (List<ColumnStatistics> sublist : Lists.partition(statisticsToSet, partitionBatchSize)) {
+            int start = counter * partitionBatchSize;
+            LOG.info("Setting column statistics for partitions {} through {}", start, start + sublist.size() - 1);
+            client.setPartitionColumnStatistics(new SetPartitionsStatsRequest(statisticsToSet));
```
abhimanyugupta07: here too. `sublist` instead of `statisticsToSet` ?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
```diff
@@ -425,18 +440,28 @@ public void alteringExistingPartitionedReplicaTableWithPartitionsSucceeds() thro
...
+    // we sent a list of 3 new partitions and used a partitionBatchSize of 2
+
+    List<List<Partition>> addCaptorValues = addPartitionCaptor.getAllValues();
+    List<Partition> firstBatch = addCaptorValues.get(0);
+    List<Partition> secondBatch = addCaptorValues.get(1);
+    assertThat(addCaptorValues.size(), is(2));
+    assertThat(firstBatch.size(), is(2));
+    assertThat(secondBatch.size(), is(1));
 
     Partition altered = alterPartitionCaptor.getValue().get(0);
```
abhimanyugupta07: I think to test the batching for `alter_partition` and `setPartitionColumnStatistics` usecase, it will be nice to have more than one existing partitions.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -30,6 +30,8 @@
 import java.util.Map;
 import java.util.Objects;
 
+import com.google.common.annotations.VisibleForTesting;
```
massdosage: Minor, but these imports should go along with the other google imports. I've made a mental note to update our hotels-oss-parent with a plugin that automatically does this that we've started using in EG org on GitHub so this is automated by Maven.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -89,6 +93,29 @@
     this.housekeepingListener = housekeepingListener;
     replicationMode = tableReplication.getReplicationMode();
     this.tableReplication = tableReplication;
+    this.partitionBatchSize = DEFAULT_PARTITION_BATCH_SIZE;
```
massdosage: NABD but you could inline this on line 77 and then only have the assignment in the constructor below which actually allows you to override the value.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -89,6 +93,29 @@
...
+  }
+
+  /**
+   * Use {@link ReplicaFactory}
```
massdosage: What does this comment mean?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -175,7 +202,13 @@ public void updateMetadata(
       if (!partitionsToCreate.isEmpty()) {
         LOG.info("Creating {} new partitions.", partitionsToCreate.size());
         try {
-          client.add_partitions(partitionsToCreate);
+          int counter = 0;
+          for (List<Partition> sublist : Lists.partition(partitionsToCreate, partitionBatchSize)) {
```
massdosage: I've never come across `Lists.partition` before, looks good.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
```diff
@@ -22,22 +22,29 @@
...
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
+import com.google.common.base.Supplier;
```
massdosage: I know it's not the most readable format but our import ordering is this: https://github.com/ExpediaGroup/eg-oss-parent/blob/master/pom.xml#L102

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
massdosage: Yeah, it feels like we are missing some tests that would have caught the issues you called out above. I'd suggest writing new tests (or modifying existing) and getting them to fail with the current code and then fixing the code and they should then pass.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
barnharts4: good catch, thanks!

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
barnharts4: yep, thanks

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
barnharts4: hmm, I agree with adding more tests for for `alter_partition` and `setPartitionColumnStatistics` but how would you suggest I get current(old) code to fail?  I guess I could mock a failure of the metastore client if the size of the list was longer than the `partitionBatchSize`, but that seems fairly contrived and not really testing anything except my ability to mock stuff.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
barnharts4: Yeah, I am struggling with the correct IntelliJ settings to match circus train's preferred import layout.  Obviously DSP standards are not matching up well with HCOM in this circumstance. Do you have a doc or a settings file or something so that I can update?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
barnharts4: It was copied from the other constructor, and it means that this constructor is package-private and to construct this, you need to use `ReplicaFactory`.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
barnharts4: ah, thanks, just what I was looking for!

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
massdosage: This will do it once I've merged and released the parent - https://github.com/HotelsDotCom/circus-train/pull/168/files

I'm getting timeouts publishing to Maven Central right now, they've had lots of scalability issues in the last few months so I'm going to leave it for today and try again tomorrow.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
barnharts4: or maybe you didn't mean the issue of overwhelming the HMS with the full list, but rather the issue that I forgot to use `sublist` above?  In that case, yes, good point.

abhimanyugupta07: Yes. Your last comment is what I meant. Just adding two more existing partitions in the test should give you a batch to trigger the changes you made.

barnharts4:
Updated test to work with any number of added and altered partitions, and any size batch.  Added tests for common edge cases and boundary conditions.  Addressed PR comments.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
massdosage: Ah, OK.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
```diff
@@ -423,41 +466,106 @@ public void alteringExistingPartitionedReplicaTableWithPartitionsSucceeds() thro
...
+      assertThat(columnStats.get(colStatNdx).getStatsDesc().getPartName(),
+              is(String.format("c=new_%s/d=new_%s_sub", addPartColStatNdx, addPartColStatNdx)));
+      assertThat(columnStats.get(colStatNdx).getStatsObj().size(), is(2));
+    }
+  }
+
+
+  @Test
+  public void alteringExistingPartitionedReplicaTableWithNewPartitionsInBatchesSucceeds_0_0() throws TException, IOException {
+    alterExistingPartitionedReplicaTableWithNewPartitionsInBatches(0,0);
```
massdosage: For another day but I wonder if we could use JUnit's "parameterised" feature to do this.

barnharts4: I looked into that, but found referenced to Parameterized not working with `MockitoJUnitRunner` (https://stackoverflow.com/questions/12606148/passing-parameterized-input-using-mockitos) so I didn't pursue further in the interest of time.  But yes, that would be the perfect for this situation.

massdosage: Ah, OK. We should move to JUnit5 at some point as then it can definitely be done, see https://github.com/ExpediaGroup/beekeeper/blob/master/beekeeper-path-scheduler-apiary/src/test/java/com/expediagroup/beekeeper/scheduler/filter/WhitelistedListenerEventFilterTest.java

PR #171, opened 2020-02-14
Upgrading Circus Train Java version to 8

massdosage:
I also took the opportunity to add a GitHub action for building circus train and made this the source for coveralls instead of Travis.


massdosage:
Fixes #171

Issue #173, opened 2020-03-04
Investigate replication of column additions in structs


patduin:
_As a user of CircusTrain_
_I'd like to know if I can change struct columns without breaking replication_

Acceptance Criteria:
* Write an integration test for the following scenario. 
  - Table (with a column of type struct) is replicated
  - column is update on source to contain extra columns. (struct type change)
  - Type change is replicated to target table correctly.

PR #174, opened 2020-03-10
Issue 173 struct evolution

max-jacobs:
Addresses #173


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/RenameTableOperation.java
```diff
@@ -0,0 +1,54 @@
...
+   * NOTE: assumes both `from` and `to` exist
+   * </p>
+   * Renames tables 'from' table into 'to' table, at the end of the operation 'from' will be gone and 'to' will be
+   * renamed.
+   */
+  public void execute(CloseableMetaStoreClient client, Table from, Table to) throws TException {
+    LOG
+        .info("Renaming table {}.{} to {}.{}", from.getDbName(), from.getTableName(), to.getDbName(),
+            to.getTableName());
+    from = client.getTable(from.getDbName(), from.getTableName());
```
patduin: Why you have to refresh the tables?

max-jacobs: The "to" table has to be refreshed I think because it doesn't have the new schema applied, only the temp table does (from the previous operation). So we could just grab the to table. But I've done both anyway for good measure. Happy to take from out.

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest.java
```diff
@@ -1398,4 +1404,115 @@ public void checkAssertion() throws Exception {
...
+    structData.put("name", "adam");
+    structData.put("city", "blackpool");
+
+    Table replicaTable = replicaHelper.createParquetPartitionedTableWithStruct(
+        toUri(replicaWarehouseUri, DATABASE, PARTITIONED_TABLE),
+        schema,
+        "struct<name:string, city:string>",
+        structData,
+        1);
+    LOG.info(">>>> Table {} ", replicaCatalog.client().getTable(DATABASE, PARTITIONED_TABLE));
```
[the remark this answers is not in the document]
abhimanyugupta07: Hmm.. I see it in all the other tests too. So, may be it has some purpose. @patduin ?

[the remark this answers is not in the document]
max-jacobs: Yeah its in all the tests, think its just for logging purposes. It's kind of handy, especially when checking a build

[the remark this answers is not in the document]
patduin: I think it's just for logging, you can leave it if you find it handy, I'm ok either way

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableService.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  public void alterTable(CloseableMetaStoreClient client, Table oldTable, Table newTable) throws TException {
+    List<FieldSchema> oldColumns = oldTable.getSd().getCols();
+    List<FieldSchema> newColumns = newTable.getSd().getCols();
+    if (hasAnyChangedColumns(oldColumns, newColumns)) {
+      LOG
+          .info("Found columns that have changed type, attempting to recreate target table with the new columns."
+              + "Old columns: {}, new columns: {}", oldColumns, newColumns);
+      Table tempTable = new Table(newTable);
```
patduin: Does this copy all the table properties?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/RenameTableOperation.java
patduin: I'm ok leaving them in that case.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/RenameTableOperation.java
```diff
@@ -0,0 +1,54 @@
...
+    String fromTableName = from.getTableName();
+    String toTableName = to.getTableName();
+    String toTableNameTemp = toTableName + "_original";
+    try {
+      from.setTableName(toTableName);
+      to.setTableName(toTableNameTemp);
+      client.alter_table(to.getDbName(), toTableName, to);
+      client.alter_table(from.getDbName(), fromTableName, from);
+    } finally {
+      client.dropTable(to.getDbName(), toTableNameTemp, false, true);
```
patduin: We need to double check beekeeper props and we don't accidentally delete folders when we drop this table.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableService.java
max-jacobs: It does indeed. We can either (1) scrub all the table params here, or (2) I've created a `DropTableService` which removes custom table params before calling the drop table command. If thats overly complicated I can remove the new service and just do (1) instead.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableService.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.thrift.TException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.hotels.hcommon.hive.metastore.client.api.CloseableMetaStoreClient;
+
+public class AlterTableService {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AlterTableService.class);
```
massdosage:
```suggestion
  private static final Logger log = LoggerFactory.getLogger(AlterTableService.class);
```
(it's not a constant)

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/ReplicaFactory.java
```diff
@@ -49,19 +55,24 @@ public ReplicaFactory(
       Supplier<CloseableMetaStoreClient> replicaMetaStoreClientSupplier,
       HousekeepingListener housekeepingListener,
       ReplicaCatalogListener replicaCatalogListener,
-      ReplicaTableFactoryProvider replicaTableFactoryPicker) {
+      ReplicaTableFactoryProvider replicaTableFactoryPicker,
```
massdosage:
```suggestion
      ReplicaTableFactoryProvider replicaTableFactoryProvider,
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/RenameTableOperation.java
```diff
@@ -0,0 +1,60 @@
...
+    LOG
+        .info("Renaming table {}.{} to {}.{}", from.getDbName(), from.getTableName(), to.getDbName(),
+            to.getTableName());
+    Table fromTable = client.getTable(from.getDbName(), from.getTableName());
+    Table toTable = client.getTable(to.getDbName(), to.getTableName());
+    String fromTableName = fromTable.getTableName();
+    String toTableName = toTable.getTableName();
+    try {
+      fromTable.setTableName(toTableName);
+      toTable.setTableName(toTableName + TEMP_SUFFIX);
```
max-jacobs: to be deleted?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableService.java
max-jacobs: Will change all these in a further PR

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/ReplicaFactory.java
max-jacobs: done :)

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableService.java
massdosage: Ack

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableService.java
massdosage: OK so if we removed the params here then we don't have to worry about the potential of Beekeeper picking up the drop of the temp table etc.? I think I prefer that approach unless some of the params are useful to have on the temp table? I don't think any of the intermediate operations use them in any way so probably not? And to be clear, the original table params will be kept on the final end result table?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+      verify(client).createTable(tempTable);
+      verify(dropTableService).removeTableParamsAndDrop(client, NEW_DB_NAME, NEW_TABLE_NAME_TEMP);
+      verify(copyPartitionsOperation).execute(client, newTable, tempTable);
+      verify(renameTableOperation).execute(client, tempTable, newTable);
+      verifyNoMoreInteractions(client);
+      assertThat(e, is(toBeThrown));
+    }
+  }
+
+  private StorageDescriptor getSd(List<FieldSchema> columns) {
```
massdosage:
```suggestion
  private StorageDescriptor createStorageDescriptor(List<FieldSchema> columns) {
```

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/hive/CopyPartitionsOperationTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+    operation.execute(client, oldTable, newTable);
+
+    verify(client).listPartitionNames(OLD_DB_NAME, OLD_TABLE_NAME, (short) -1);
+    verify(client).getPartitionsByNames(OLD_DB_NAME, OLD_TABLE_NAME, Collections.emptyList());
+    verifyNoMoreInteractions(client);
+  }
+
+  private List<Partition> createPartitions(int count) {
+    ArrayList<Partition> partitions = new ArrayList<>();
```
massdosage:
```suggestion
    List<Partition> partitions = new ArrayList<>();
```

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableServiceTest.java
```diff
@@ -0,0 +1,135 @@
...
+    service.removeTableParamsAndDrop(client, DB_NAME, TABLE_NAME);
+
+    verify(client).getTable(DB_NAME, TABLE_NAME);
+    verify(client).dropTable(DB_NAME, TABLE_NAME, false, true);
+    verifyNoMoreInteractions(client);
+  }
+
+  @Test
+  public void removeParamsAndDrop() throws TException {
+    HashMap<String, String> params = new HashMap<>();
```
massdosage:
```suggestion
    Map<String, String> params = new HashMap<>();
```

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableServiceTest.java
```diff
@@ -0,0 +1,135 @@
...
+    List<Table> capturedTables = tableCaptor.getAllValues();
+    assertThat(capturedTables.size(), is(1));
+    Map<String, String> parameters = capturedTables.get(0).getParameters();
+    assertThat(parameters.size(), is(1));
+    assertThat(parameters.get("EXTERNAL"), is("true"));
+  }
+
+  @Test
+  public void removeParamsAndDropCaseInsensitiveExternalTable() throws TException {
+    HashMap<String, String> params = new HashMap<>();
```
massdosage:
```suggestion
    Map<String, String> params = new HashMap<>();
```

circus-train-integration-tests/pom.xml
```diff
@@ -172,6 +165,28 @@
...
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-avro</artifactId>
+      <version>1.10.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.avro</groupId>
+      <artifactId>avro</artifactId>
+      <version>1.9.0</version>
+    </dependency>
+    <dependency>
```
massdosage: Why did you move this down here? Ideally the dependencies should be alpha sorted by groupId and then artifactId in two sections - one for normal scope and another for test scope.

circus-train-integration-tests/pom.xml
```diff
@@ -172,6 +165,28 @@
       <artifactId>protobuf-java</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
massdosage: I'd be wary of setting these to version numbers other than what the Hive/Hadoop dependencies are using. I guess this works but I just know how fragile things can get as the newer versions aren't backwards compatible and the only reason we're not seeing an issue now is we haven't hit certain code paths that would trigger problems.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/partitioned-single-table-one-partition.yml
```diff
@@ -0,0 +1,15 @@
...
+      partition-limit: 100
+    replica-table:
+      table-location: ${circus-train-runner.replica-warehouse-uri}/${circus-train-runner.database-name}/ct_table_p
+    orphaned-data-strategy: NONE
+copier-options:
+  file-attribute: replication, blocksize, user, group, permission, checksumtype
+  preserve-raw-xattrs: false
+metrics-reporter:
+  period: 1
+  time-unit: SECONDS
```
massdosage: Line break

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest.java
```diff
@@ -1398,4 +1404,119 @@ public void checkAssertion() throws Exception {
...
+        .type()
+        .record("details_struct")
+        .fields()
+        .requiredString("name")
+        .requiredString("city")
+        .endRecord()
+        .noDefault()
+        .endRecord();
+
+    HashMap<String, String> structData = new HashMap<>();
```
massdosage:
```suggestion
    Map<String, String> structData = new HashMap<>();
```

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/IntegrationTestHelper.java
```diff
@@ -74,6 +87,60 @@ void createPartitionedTable(URI sourceTableUri) throws Exception {
...
+    Path filePath = new Path(partitionFile.toURI());
+    ParquetWriter<GenericData.Record> writer = AvroParquetWriter.<GenericData.Record>builder(filePath)
+        .withSchema(schema)
+        .withConf(new Configuration())
+        .build();
+
+    try {
+      writer.write(record);
+    } catch (IOException e) {
+      e.printStackTrace();
```
massdosage: Don't we want to throw this and fail the test if this happens?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableService.java
max-jacobs: Okay cool, I'm happy to do this, I'll remove the DropTableService, and just remove the params as part of this service.

The intermediate ops dont use the params, but i think beekeeper may pick up on the alter partition events (though will ignore them as the location doesnt change). And yes, the original params will be on the final table - i'll add some tests for this.

max-jacobs: Tried this out, and actually it doesn't seem possible. We can't do an alter table on the table as it still has the old schema, so it fails with the same error. So the DropTableService looks better as we do the alter/drop once everything has been updated.

circus-train-integration-tests/pom.xml
max-jacobs: Think this was out of desperation initially to get the parquet files writing successfully. Will revert.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/AlterTableService.java
massdosage: OK, let's go with that for now and keep an eye on it when this is in use.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/hive/RenameTableOperationTest.java
```diff
@@ -0,0 +1,118 @@
...
+    operation.execute(client, fromTable, toTable);
+
+    fromTable.setTableName(TO_TABLE_NAME);
+    verify(client).alter_table(TO_DB_NAME, TO_TABLE_NAME, toTableTemp);
+    verify(client).alter_table(FROM_DB_NAME, FROM_TABLE_NAME, fromTable);
+    verify(dropTableService).removeTableParamsAndDrop(client, TO_DB_NAME, TO_TABLE_NAME_TEMP);
+  }
+
+  @Test
+  public void renameTableExceptionThrown1() throws Exception {
```
massdosage: NABD but would be nice to replace "1" and "2" in this and the next method with something more descriptive as to what the difference is. From what I can see it's about where the exception is thrown, when altering the to or the from table? So maybe something like `renameToTableException` and `renameFromTableException`?

PR #182, opened 2020-03-31
Disable coveralls for forks

massdosage:
This change only runs Coveralls for pull requests on non-forked branches. This is because secrets (like the Coveralls repo token) aren't allowed to be shared outside of the org (makes sense) but this was previously failing the build right at the end when trying to send the report to Coveralls as the token for that was missing. So instead what we do is disable the Coveralls report if the secret is missing. I've tested this from a fork and non-fork and it seems to work correctly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,9 +10,22 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
+    - name: Init Coveralls
+      shell: bash
+      run: |
+          COVERALLS_TOKEN=${{ secrets.COVERALLS_REPO_TOKEN }}
+          if [[ -z "${COVERALLS_TOKEN}" ]];
+          then
+             echo "Coveralls token not available"
+             COVERALLS_SKIP=true
+          else
+             echo "Coveralls token available"
+             COVERALLS_SKIP=false
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -10,9 +10,22 @@ jobs:
...
+          COVERALLS_TOKEN=${{ secrets.COVERALLS_REPO_TOKEN }}
+          if [[ -z "${COVERALLS_TOKEN}" ]];
+          then
+             echo "Coveralls token not available"
+             COVERALLS_SKIP=true
+          else
+             echo "Coveralls token available"
+             COVERALLS_SKIP=false
+          fi
+          echo ::set-env name=COVERALLS_SKIP::${COVERALLS_SKIP}
```
KenFigueiredo: Was a bit unfamiliar with how github handled passing dynamic variables in between build steps in a workflow. Interesting that they went this log syntax: https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

massdosage: Yes, so was I. I initially tried `set-output` but couldn't get it working so went with `set-env`, it doesn't make much difference in this case. It is an "interesting" syntax indeed, I don't think I'm a fan but maybe I'll come around to it.

PR #184, opened 2020-04-23
Upgrade Hive to 2.3.7

vedantchokshi:
### :pencil: Description
* Changed version of `hive.version` to `2.3.7` (was `2.3.2`).

### :link: Related Issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [16.1.1] - TBD
+### Changed
+* Changed version of `hive.version` to `2.3.7` (was `2.3.2`).
+
 ## [16.1.0] - 2020-03-18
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [16.1.1] - TBD
+### Changed
+* Changed version of `hive.version` to `2.3.7` (was `2.3.2`).
```
massdosage, line 3:
```suggestion
* Changed version of `hive.version` to `2.3.7` (was `2.3.2`). This allows Circus Train to be used on JDK>=9.
```
(let's add an extra bit of detail here so public end users can see what the main benefit of this change is).

Issue #183, opened 2020-04-17
Support for replication after non-backwards-compatible schema changes


Dima1224:
**Is your feature request related to a problem? Please describe.**
Early on in the development lifecycle, it is common to change a table's schema in non-backwards-compatible ways and repopulate the table from scratch. This could also happen with a mature table, though it would be much more rare.

**Describe the solution you'd like**
I'd like Circus Train to support this use case and replicate the update. As a user of Circus Train I am not expecting it to protect me from inadvertent breaking changes, I just expect it to copy data and metadata from one place to the next.

If there are CT users who depend on CT to protect them from breaking schema changes, I propose we add some metadata to indicate which tables should be protected and which shouldn't. This can be thought of as a dev/prod distinction or something along the lines of safe/unsafe. CT could still attempt to detect partitions which weren't updated to match the schema change and fail to copy if such partitions exist. Though I'd argue that this isn't the job of the copy tool, but rather the Data Lake tooling surrounding the upstream schema change.

This has been discussed with @massdosage

JayGreeeen:
Thanks for the extension suggestion! 

I'm working on adding in a `FULL_OVERWRITE` replication mode, which when specified in the config file will drop the existing target table and replace it with a copy of the source table, keeping the same name as the dropped target table. Effectively, doing a full update of data and metadata each time.
Is this what you're after?

Dima1224:
Yep, that sounds like it would do the trick. One thing I'd keep in mind is that this will likely be the mode for tables that are being actively iterated on. At some point that table will be hardened and the consumer will likely want to switch the replication mode to "regular."

JayGreeeen:
Great! And yep thats fine, when you no longer want to overwrite the target each time you can just use the `FULL` replication mode as done previously.

JayGreeeen:
The PRs relating to this issue have both been merged, I will now close this ticket.

PR #185, opened 2020-05-12
Full overwrite replication

JayGreeeen:
### :pencil: Description
Added a replication mode `FULL_OVERWRITE` which will overwrite an existing table with the source. The expected use for this mode would be early during the development cycle when the schema could be subject to incompatible changes. 

### :link: Related Issues
#183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 ### Changed
 * Changed version of `hive.version` to `2.3.7` (was `2.3.2`). This allows Circus Train to be used on JDK>=9.
 
+### Added
+* Replication mode `FULL_OVERWRITE` to overwrite a previously replicated table. Useful for incompatible schema changes. 
+
 ## [16.1.0] - 2020-03-18
 ### Changed
 * Updated S3S3Copier to have a configurable max number of threads to pass to TransferManager.
--- a/circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/conf/ReplicationMode.java
+++ b/circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/conf/ReplicationMode.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2019 Expedia, Inc.
+ * Copyright (C) 2016-2020 Expedia, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,8 @@ public enum ReplicationMode {
 
   FULL,
   METADATA_MIRROR,
-  METADATA_UPDATE;
+  METADATA_UPDATE,
+  FULL_OVERWRITE;
+
 
 }
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
@@ -107,6 +107,8 @@ public class ReplicationFactoryImpl implements ReplicationFactory {
       replication = new PartitionedTableMetadataMirrorReplication(sourceDatabaseName, sourceTableName,
           partitionPredicate, source, replica, eventIdFactory, replicaDatabaseName, replicaTableName);
       break;
+
+    case FULL_OVERWRITE:
     case FULL:
       Map<String, Object> mergedCopierOptions = tableReplication
           .getMergedCopierOptions(copierOptions.getCopierOptions());
@@ -141,6 +143,7 @@ public class ReplicationFactoryImpl implements ReplicationFactory {
       replication = new UnpartitionedTableMetadataMirrorReplication(sourceDatabaseName, sourceTableName, source,
           replica, eventIdFactory, replicaDatabaseName, replicaTableName);
       break;
+    case FULL_OVERWRITE:
     case FULL:
       Map<String, Object> mergedCopierOptions = tableReplication
           .getMergedCopierOptions(copierOptions.getCopierOptions());
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
@@ -107,7 +107,6 @@ public class ReplicationFactoryImpl implements ReplicationFactory {
       replication = new PartitionedTableMetadataMirrorReplication(sourceDatabaseName, sourceTableName,
           partitionPredicate, source, replica, eventIdFactory, replicaDatabaseName, replicaTableName);
       break;
-
     case FULL_OVERWRITE:
     case FULL:
       Map<String, Object> mergedCopierOptions = tableReplication
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Strings.nullToEmpty;
 import static com.hotels.bdp.circustrain.api.CircusTrainTableParameter.REPLICATION_EVENT;
 import static com.hotels.bdp.circustrain.api.CircusTrainTableParameter.REPLICATION_MODE;
 import static com.hotels.bdp.circustrain.api.conf.ReplicationMode.FULL;
+import static com.hotels.bdp.circustrain.api.conf.ReplicationMode.FULL_OVERWRITE;
 import static com.hotels.bdp.circustrain.api.conf.ReplicationMode.METADATA_MIRROR;
 import static com.hotels.bdp.circustrain.api.conf.ReplicationMode.METADATA_UPDATE;
 import static com.hotels.hcommon.hive.metastore.util.LocationUtils.locationAsPath;
@@ -103,15 +104,15 @@ public class Replica extends HiveEndpoint {
    */
   @VisibleForTesting
   Replica(
-          ReplicaCatalog replicaCatalog,
-          HiveConf replicaHiveConf,
-          Supplier<CloseableMetaStoreClient> replicaMetaStoreClientSupplier,
-          ReplicaTableFactory replicaTableFactory,
-          HousekeepingListener housekeepingListener,
-          ReplicaCatalogListener replicaCatalogListener,
-          TableReplication tableReplication,
-          AlterTableService alterTableService,
-          int partitionBatchSize) {
+      ReplicaCatalog replicaCatalog,
+      HiveConf replicaHiveConf,
+      Supplier<CloseableMetaStoreClient> replicaMetaStoreClientSupplier,
+      ReplicaTableFactory replicaTableFactory,
+      HousekeepingListener housekeepingListener,
+      ReplicaCatalogListener replicaCatalogListener,
+      TableReplication tableReplication,
+      AlterTableService alterTableService,
+      int partitionBatchSize) {
     super(replicaCatalog.getName(), replicaHiveConf, replicaMetaStoreClientSupplier);
     this.replicaCatalogListener = replicaCatalogListener;
     tableFactory = replicaTableFactory;
@@ -302,7 +303,14 @@ public class Replica extends HiveEndpoint {
     LOG.info("Updating replica table metadata.");
     TableAndStatistics replicaTable = tableFactory
         .newReplicaTable(eventId, sourceTable, replicaDatabaseName, replicaTableName, tableLocation, replicationMode);
+
+    // remove table here so the below doesnt give a table ?
+    if (replicationMode == FULL_OVERWRITE) {
+      dropReplicaTable(client, replicaDatabaseName, replicaTableName);
+    }
+
     Optional<Table> oldReplicaTable = getTable(client, replicaDatabaseName, replicaTableName);
+
     if (!oldReplicaTable.isPresent()) {
       LOG.debug("No existing replica table found, creating.");
       try {
@@ -312,6 +320,7 @@ public class Replica extends HiveEndpoint {
         throw new MetaStoreClientException(
             "Unable to create replica table '" + replicaDatabaseName + "." + replicaTableName + "'", e);
       }
+
     } else {
       makeSureCanReplicate(oldReplicaTable.get(), replicaTable.getTable());
       LOG.debug("Existing replica table found, altering.");
@@ -323,6 +332,7 @@ public class Replica extends HiveEndpoint {
             "Unable to alter replica table '" + replicaDatabaseName + "." + replicaTableName + "'", e);
       }
     }
+
     return oldReplicaTable;
   }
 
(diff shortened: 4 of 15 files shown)


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
```diff
@@ -107,6 +107,7 @@ private Replication createPartitionedTableReplication(
       replication = new PartitionedTableMetadataMirrorReplication(sourceDatabaseName, sourceTableName,
           partitionPredicate, source, replica, eventIdFactory, replicaDatabaseName, replicaTableName);
       break;
+    case FULL_OVERWRITE:
```
max-jacobs, line 110: I think this and the added case in the other switch statement needs some tests

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -426,4 +436,25 @@ public TableAndStatistics getTableAndStatistics(TableReplication tableReplicatio
     return super.getTableAndStatistics(tableReplication.getReplicaDatabaseName(),
         tableReplication.getReplicaTableName());
   }
+
+  private void dropReplicaTable(CloseableMetaStoreClient client, String replicaDatabaseName, String replicaTableName) {
```
max-jacobs, line 440: Would recommend moving this method to the `DropTableService`. Also, on a full overwrite I imagine we will want to delete the current data (right?), but if we're using Beekeeper drop table events will be ignored (unless we add the param to allow drop table events). But even if we do this, we won't want to schedule the table path because then Beekeeper will delete any new data that is added after the delete - Beekeeper doesn't yet go to the Metastore to only schedule the current partitions for deletion on a table drop.

We probably want to do an alter all the individual partitions to make sure they're deleted and not the new data? Not sure, is that possible?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
```diff
@@ -323,6 +327,60 @@ public void validateReplicaTableMetadataMirrorOnExistingMetadataUpdateTableFails
...
+    try {
+    when(mockMetaStoreClient.tableExists(DB_NAME, TABLE_NAME)).thenReturn(false);
+
+    tableReplication.setReplicationMode(FULL_OVERWRITE);
+    replica
+        .updateMetadata(EVENT_ID, tableAndStatistics,
+            new PartitionsAndStatistics(sourceTable.getPartitionKeys(), Collections.<Partition>emptyList(),
+                Collections.<String, List<ColumnStatisticsObj>>emptyMap()),
+            DB_NAME, TABLE_NAME, mockReplicaLocationManager);
+
```
max-jacobs, line 375: think you need a fail here to make sure this throws an exception

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/partitioned-single-table-full-overwrite.yml
```diff
@@ -0,0 +1,9 @@
+table-replications:
```
max-jacobs, line 1: I think at some point we should think about removing all these duplicated yaml files and building them programmatically, just an idea

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
max-jacobs: Spoke with the team, we think that this is true for unpartitioned tables, but for partitioned tables the table path doesn't include the event id, the partition paths do instead. So we can't do a drop without first making sure that Beekeeper isn't listening otherwise we'll lose all the new data too.

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest.java
```diff
@@ -850,6 +851,123 @@ public void checkAssertion() throws Exception {
...
+  }
+
+  private void setupReplicaTable(Table replicaTable, boolean partitioned, URI replicaLocation) throws Exception {
+    setupRelicaParameters(replicaTable);
+    if (partitioned) {
+      addPartitionsToReplica(replicaTable, replicaLocation);
+    }
+  }
+
+  private void setupRelicaParameters(Table replicaTable) {
```
abhimanyugupta07, line 953:
```suggestion
  private void setupReplicaParameters(Table replicaTable) {
```

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest.java
```diff
@@ -850,6 +851,123 @@ public void checkAssertion() throws Exception {
...
+        assertThat(hiveTable.getParameters().get("paramToUpdate"), is("updated"));
+        assertThat(isExternalTable(hiveTable), is(true));
+        assertThat(hiveTable.getSd().getCols(), is(DATA_COLUMNS));
+      }
+    });
+    runner.run(config.getAbsolutePath());
+  }
+
+  private void setupReplicaTable(Table replicaTable, boolean partitioned, URI replicaLocation) throws Exception {
+    setupRelicaParameters(replicaTable);
```
abhimanyugupta07, line 947:
```suggestion
    setupReplicaParameters(replicaTable);
```

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest.java
abhimanyugupta07, line 953: This function is not just setting params but also adding two new columns to the table. Splitting this function up would be better.

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainS3S3IntegrationTest.java
```diff
@@ -273,4 +282,139 @@ public void checkAssertion() throws Exception {
...
+  }
+
+  private void setupReplicaTable(Table replicaTable, boolean partitioned, URI replicaLocation) throws Exception {
+    setupRelicaParameters(replicaTable);
+    if (partitioned) {
+      addPartitionsToReplica(replicaTable, replicaLocation);
+    }
+  }
+
+  private void setupRelicaParameters(Table replicaTable) {
```
abhimanyugupta07, line 402: same comments here. Typo and splitting up.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 ### Changed
 * Changed version of `hive.version` to `2.3.7` (was `2.3.2`). This allows Circus Train to be used on JDK>=9.
 
+### Added
+* Replication mode `FULL_OVERWRITE` to overwrite a previously replicated table. Useful for incompatible schema changes.
```
massdosage, line 6: Let's go with `16.2.0` as the version number as this is a new feature.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -346,12 +351,12 @@ public void validateReplicaTable(String replicaDatabaseName, String replicaTable
       Optional<Table> oldReplicaTable = getTable(client, replicaDatabaseName, replicaTableName);
       if (oldReplicaTable.isPresent()) {
         LOG.debug("Existing table found, checking that it is a valid replica.");
-        determinValidityOfReplica(replicationMode, oldReplicaTable.get());
+        determineValidityOfReplica(replicationMode, oldReplicaTable.get());
```
massdosage, line 354: Good spot ;)

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/conf/ReplicationMode.java
```diff
@@ -19,6 +19,8 @@
 
   FULL,
   METADATA_MIRROR,
-  METADATA_UPDATE;
+  METADATA_UPDATE,
+  FULL_OVERWRITE;
```
massdosage, line 23: Move this up to below `FULL` - they feel more "related" and I don't think we use the ordering of this enum for anything.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/partitioned-single-table-full-overwrite.yml
massdosage: The intention initially was for the integration tests to read in the yaml to ensure the yaml conversion code etc. is working. I don't think we envisioned there being this many integration tests, ideally we should only add them to test things that can't be tested in any other way.

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainS3S3IntegrationTest/unpartitioned-single-table-full-overwrite-replication.yml
```diff
@@ -0,0 +1,10 @@
+table-replications:
```
massdosage, line 1: Having looked through the above I'm not sure we need to add these tests to all combinations of HDFS and S3 as they probably don't really test anything new in each of them right? Perhaps create a new `CircusTrainReplicationModeIntegrationTest` and then put the two tests into that and have it use S3 to S3 for the location on each side (as that's becoming the most used file system on both sides now). 

At some point we should review the other integration tests and see if there is other overlap that isn't needed that can be refactored out to reduce the number of these.

JayGreeeen: Yeah it felt pretty repetitive, this sounds like a better idea 👍

circus-train-integration-tests/src/test/data/com/hotels/bdp/circustrain/integration/CircusTrainHdfsHdfsIntegrationTest/partitioned-single-table-full-overwrite.yml
max-jacobs: Yep understood, we could have a couple of template yaml files, and create temp files from these for each test on the fly

JayGreeeen: Adrian suggested that I add an integration test class for the replication modes, so that will remove all these duplicated files 👍

massdosage: I think a separate task could be to evaluate the existing integration tests, clean them up so we have the minimal necessary and we could look into templated files as part of that. I'd suggest a separate PR though.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -426,4 +436,25 @@ public TableAndStatistics getTableAndStatistics(TableReplication tableReplicatio
     return super.getTableAndStatistics(tableReplication.getReplicaDatabaseName(),
         tableReplication.getReplicaTableName());
   }
+
+  private void dropReplicaTable(CloseableMetaStoreClient client, String replicaDatabaseName, String replicaTableName) {
+    LOG.debug("Replication mode: FULL_OVERWRITE. Dropping existing replica table.");
+    try {
+      if (client.tableExists(replicaDatabaseName, replicaTableName)) {
```
patduin, line 443: This call is a "new" addition of the hive API I'm not sure if we use it anywhere else but if we do it is ok but if we don't better to avoid using it as it's not always there on old client.
Better to just do the 
```
Try
dropTable() 
Catch(NoSuchObjectException e) {
//nothing to drop table doesn't exist etc.....
}
```

JayGreeeen:
We've decided that we _do_ want this replication mode to delete the data as well as the replica table, but this is a slightly bigger change so we will separate it out as another PR. 

So for now, this mode will only drop the existing replica table and not touch any data.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -302,6 +304,16 @@ public void updateMetadata(
...
     TableAndStatistics replicaTable = tableFactory
         .newReplicaTable(eventId, sourceTable, replicaDatabaseName, replicaTableName, tableLocation, replicationMode);
+
+    if (replicationMode == FULL_OVERWRITE) {
+      LOG.debug("Replication mode: FULL_OVERWRITE. Dropping existing replica table and its data.");
+      DropTableService dropTableService = new DropTableService();
+      try {
+        dropTableService.removeTableParamsAndDrop(client, replicaDatabaseName, replicaTableName);
+      } catch (TException e) {
+        LOG.info("No replica table '" + replicaDatabaseName + "." + replicaTableName + "' found. Nothing to delete.");
```
massdosage, line 314: I think the `TException` could be thrown for other reasons too, like network problems etc. So perhaps it would be better to have a try/catch for `TException` that retrhrows a `MetaStoreClientException` like in some of the blocks below, and then introduce a "does table exist" check in the code (maybe in the service?) and if it doesn't then it just silently continues without throwing an exception (as this is fine, there's no need to drop the replica if it doesn't exist).

patduin: NoSuchObjectException is the one that get's throw if the table/db doesn't exist

massdosage: OK, so we could have a more specific catch block for that here, or do we want to change the service to not throw that and silently continue? Do we have other cases that use it where we want that exception thrown?

JayGreeeen: Yeah I also realised this. I have fixed it in the other branch I have for deleting data so its now:
```
} catch (Exception e) {
   LOG.info("Replica table '" + replicaDatabaseName + "." + replicaTableName + "' was not dropped.");
}
```

And in the drop table service it catches the NoSuchObj exception and logs that the table wasnt found.

massdosage: Do you want to update this branch too? The plan is to merge the branches in separately right? First this one to get the "drop metadata" changes in and then the other one gets the "drop file data" changes.

JayGreeeen: Yeah thats the plan. Will update this one with those logs 👍

patduin: I think it is ok that the DropTableService swallows the NoSuchObjectE. But then I agree with Adrian this block should catch and rethrow MetaStoreClientException. Don't log anything that's not up to this class to silently ignore this.

JayGreeeen: Thats a good point. Will log the NoSuchObj inside the dropTableService, and here will throw a MetaStoreClientExc in the catch.

PR #186, opened 2020-05-14
Added README for resolving issue with integration tests

massdosage:
Based on similar README - https://github.com/HotelsDotCom/waggle-dance/blob/master/waggle-dance-integration-tests/README.md

### DIFF
--- a/circus-train-integration-tests/README.md
+++ b/circus-train-integration-tests/README.md
@@ -19,9 +19,9 @@ Caused by: java.lang.ClassNotFoundException: com.google.common.base.MoreObjects$
 This indicates that the shading and relocation (in the above example of a Guava class) hasn't been activated in the IDE. One solution to this is to 
 rebuild the entire project using Maven on the command line and then re-run the tests in the IDE. The following steps should achieve that: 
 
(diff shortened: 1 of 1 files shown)


circus-train-integration-tests/README.md
```diff
@@ -0,0 +1,27 @@
...
+	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
+	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
+	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
+	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)    
+```
+This indicates that the shading and relocation (in the above example of a Guava class) hasn't been activated in the IDE. One solution to this is to 
+rebuild the entire project using Maven on the command line and then re-run the tests in the IDE. The following steps should achieve that: 
+
+1. Run `mvn -DskipTests install` in the top-level Circus Train parent project (this should repackage and relocate classes using the Maven 
+plugin and generate the correct class files in the produce jar artifacts).
```
barnharts4, line 23: Not sure what `in the produce jar artifacts` means.  Did you mean `in the produced jar artifacts`?

massdosage: Yes I did, "packaged" is probably a better word, will change it.

PR #187, opened 2020-06-10
Full overwrite replication delete data

JayGreeeen:
### :pencil: Description
PR for deleting data associated with the existing replica table when using `FULL_OVERWRITE` replication.
I have added new DataManipulationClients which are clients with the correct access for respective replications - `s3->s3`, `hdfs->s3`, `hdfs->hdfs`. 

### :link: Related Issues
#185 - this is the PR to implement the new replication mode, currently only drops the existing table.


circus-train-distcp-copier/src/main/java/com/hotels/bdp/circustrain/distcpcopier/HdfsDataManipulationClientFactory.java
```diff
@@ -43,22 +43,22 @@ public HdfsDataManipulationClientFactory(@Value("#{replicaHiveConf}") Configurat
...
+  // The HDFS client doesn't need to use the path.
   @Override
   public DataManipulationClient newInstance(String path) {
     return new HdfsDataManipulationClient(conf);
   }
 
   /**
-   * Checks that the replica location is are an hdfs location. This will delete replica data whether it is been
-   * replicated from s3 or hdfs.
+   * Checks that the replica location is an hdfs location. This will delete replica data whether it is been replicated
```
cmathiesen, line 53:
```suggestion
   * Checks that the replica location is an hdfs location. This will delete replica data whether it has been replicated
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulationClientFactoryManager.java
```diff
@@ -0,0 +1,82 @@
...
+  @PostConstruct
+  void postConstruct() {
+    LOG.debug("Initialized with {} DataManipulationClientFactories", clientFactories.size());
+    for (DataManipulationClientFactory clientFactory : clientFactories) {
+      LOG.debug("ClientFactory class {}", clientFactory.getClass().getName());
+    }
+  }
+
+  public DataManipulationClient getClientForPath(String replicaLocation) {
+    for (DataManipulationClientFactory clientFactory : clientFactories) {
```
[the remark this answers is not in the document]
JayGreeeen: yeah thats a good point. Might be a good idea to include that then

README.md
```diff
@@ -328,6 +328,7 @@ If data is being replicated to HDFS then Circus Train will use DistCp to copy th
             skip-crc: false
             ssl-configuration-file: /foo/bar/ssl-config
             copier-factory-class: com.hotels.bdp.circustrain.distcpcopier.DistCpCopier
+            client-manipulation-factory-class: com.hotels.bdp.circustrain.distcpcopier.HdfsDataManipulationClientFactory
```
massdosage, line 331: Having looked through this I'm not sure whether we need to expose the term "client". Could we just call it a Data Manipulator? Feels more generic to me. Then this property would become `data-manipulator-factory-class`. Obviously impacts on the rest of the doc so I won't comment again on this below. What do others think?

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/data/DataManipulationClient.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.api.data;
+
+import java.io.IOException;
+
+public interface DataManipulationClient {
```
massdosage, line 20:
```suggestion
public interface DataManipulator {
```

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/data/DataManipulationClient.java
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * Returns boolean stating whether the deletion was successful or not.
```
massdosage, line 23: Add javadoc describing when the IOE could occur. As an end user I'd be most curious to know what the difference is between this returning `false` and an exception occurring,

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/data/DataManipulationClientFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.api.data;
+
+import java.util.Map;
+
+import org.apache.hadoop.fs.Path;
+
+public interface DataManipulationClientFactory {
```
massdosage, line 22:
```suggestion
public interface DataManipulatorFactory {
```

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/data/DataManipulationClientFactoryManager.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.api.data;
+
+import java.util.Map;
+
+import org.apache.hadoop.fs.Path;
+
+public interface DataManipulationClientFactoryManager {
```
massdosage, line 22:
```suggestion
public interface DataManipulatorFactoryManager {
```

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AwsDataManipulationClient.java
```diff
@@ -0,0 +1,83 @@
...
+import com.amazonaws.services.s3.model.DeleteObjectsRequest.KeyVersion;
+import com.amazonaws.services.s3.model.DeleteObjectsResult;
+import com.amazonaws.services.s3.model.ObjectListing;
+import com.amazonaws.services.s3.model.S3ObjectSummary;
+
+import com.hotels.bdp.circustrain.api.data.DataManipulationClient;
+
+public class AwsDataManipulationClient implements DataManipulationClient {
+
+  private static final Logger LOG = LoggerFactory.getLogger(AwsDataManipulationClient.class);
```
massdosage, line 40:
```suggestion
  private static final Logger log = LoggerFactory.getLogger(AwsDataManipulationClient.class);
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulationClientFactoryManager.java
```diff
@@ -0,0 +1,89 @@
...
+import com.hotels.bdp.circustrain.api.Modules;
+import com.hotels.bdp.circustrain.api.data.DataManipulationClientFactory;
+import com.hotels.bdp.circustrain.api.data.DataManipulationClientFactoryManager;
+
+@Profile({ Modules.REPLICATION })
+@Component
+@Order(Ordered.LOWEST_PRECEDENCE)
+public class DefaultDataManipulationClientFactoryManager implements DataManipulationClientFactoryManager {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultDataManipulationClientFactoryManager.class);
```
massdosage, line 41:
```suggestion
  private static final Logger log = LoggerFactory.getLogger(DefaultDataManipulationClientFactoryManager.class);
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -445,4 +437,23 @@ public TableAndStatistics getTableAndStatistics(TableReplication tableReplicatio
         tableReplication.getReplicaTableName());
   }
 
+  public void checkIfReplicaCleanupRequired(
```
massdosage, line 440: This does more than just check, it also performs the cleanup. Perhaps rename to `cleanupReplicaIfRequired`. Or maybe better move the `if (replicationMode == FULL_OVERWRITE) {` up to where it was before and then call this method `cleanupReplica` as it only applies to that mode.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -445,4 +437,23 @@ public TableAndStatistics getTableAndStatistics(TableReplication tableReplicatio
...
+      String replicaDatabaseName,
+      String replicaTableName,
+      DataManipulationClient dataManipulationClient) {
+    try (CloseableMetaStoreClient client = getMetaStoreClientSupplier().get()) {
```
massdosage, line 444: Wouldn't it be better to move this to line 448? There's no need to create a client unless we get that far.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -445,4 +437,23 @@ public TableAndStatistics getTableAndStatistics(TableReplication tableReplicatio
...
+      if (replicationMode == FULL_OVERWRITE) {
+        LOG.debug("Replication mode: FULL_OVERWRITE. Checking for existing replica table.");
+        DropTableService dropTableService = new DropTableService();
+        try {
+            dropTableService
+              .dropTableAndData(client, replicaDatabaseName, replicaTableName,
+                  dataManipulationClient);
+        } catch (Exception e) {
+          LOG.info("Replica table '" + replicaDatabaseName + "." + replicaTableName + "' was not dropped.");
```
massdosage, line 453: Should we log `e` here too so we know why?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -25,28 +28,69 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(DropTableService.class);
   private static final String EXTERNAL_KEY = "EXTERNAL";
   private static final String IS_EXTERNAL = "TRUE";
+  private static final short BATCH_SIZE = (short) 10;
 
   /**
    * Removes all parameters from a table before dropping the table.
    */
-  public void removeTableParamsAndDrop(
+  public void removeTableParamsAndDrop(CloseableMetaStoreClient client, String databaseName, String tableName)
```
massdosage, line 45: How is this different to `dropTable` now? If we're always going to remove the params in that method then I'm not sure we need this one anymore?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -25,28 +28,69 @@
...
+    throws TException {
+    LOG.debug("Dropping table {}.{} and its data.", databaseName, tableName);
+    Table table = getTable(client, databaseName, tableName);
+
+    if (table != null) {
+      String replicaLocation = table.getSd().getLocation();
+      if (table.getPartitionKeysSize() == 0) {
+        deleteData(dataManipulationClient, replicaLocation);
+      } else {
+        List<String> partitionLocations = getPartitionLocations(client, table);
```
massdosage, line 71: Are we OK with not doing the whole partition iterator trick here? I guess for most of this type of table they're generally not very big so it's unlikely this will return tens of thousands of partitions and run out of memory like we've had in other parts of the code so it's probably fine.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -56,13 +100,52 @@ public void removeTableParamsAndDrop(
...
-    LOG
-        .info("Dropping table '{}.{}'.", table.getDbName(), table.getTableName());
-    client.dropTable(table.getDbName(), table.getTableName(), false, true);
+    LOG.info("Dropping table '{}.{}'.", databaseName, tableName);
+    client.dropTable(databaseName, tableName, false, true);
+  }
+
+  private void deleteData(DataManipulationClient dataManipulationClient, String replicaDataLocation) {
+    try {
+      LOG.info("Dropping table data from location: {}.", replicaDataLocation);
```
massdosage, line 109:
```suggestion
      LOG.info("Deleting table data from location: {}.", replicaDataLocation);
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -56,13 +100,52 @@ public void removeTableParamsAndDrop(
...
+      boolean dataDeleted = dataManipulationClient.delete(replicaDataLocation);
+      LOG.info("Data deleted: {}.", dataDeleted);
+    } catch (IOException e) {
+      LOG.info("Could not drop replica table data at location:{}.", replicaDataLocation);
```
massdosage, line 113:
```suggestion
      LOG.info("Error deleting table data from location:{}.", replicaDataLocation);
```
I think we should also log `e` so we can see why it failed to delete the data.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -56,13 +100,52 @@ public void removeTableParamsAndDrop(
...
+    }
+  }
+
+  private void deletePartitionData(
+      DataManipulationClient dataManipulationClient,
+      String replicaTableLocation,
+      List<String> replicaPartitionLocations) {
+    try {
+      LOG.info("Dropping partition data from base location: {}.", replicaTableLocation);
```
massdosage, line 122:
```suggestion
      LOG.info("Deleting partition data from base location: {}.", replicaTableLocation);
```
(although, I'm not sure it's even worth logging or using this as theoretically the partition locations could be completely unrelated to the table's base location.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -56,13 +100,52 @@ public void removeTableParamsAndDrop(
...
+      for (String location : replicaPartitionLocations) {
+        boolean deleted = dataManipulationClient.delete(location);
+        LOG.debug("Attempted to delete data from location: {}. Successful deletion = {}.", location, deleted);
+      }
+    } catch (IOException e) {
```
massdosage, line 127: At the moment this will stop if one of the partitions couldn't be dropped. Wouldn't it be better to put the try/catch around line 124 so it can try each location? And then basically do an AND on the booleans returned so if any one is false then the method returns false.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -56,13 +100,52 @@ public void removeTableParamsAndDrop(
...
+      LOG.info("Could not drop replica partition data at location:{}.", replicaTableLocation);
```
massdosage, line 128: Then this message can log the actual partition location which will be more meaningful (and also log the exception)

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -56,13 +100,52 @@ public void removeTableParamsAndDrop(
...
+    } catch (UnsupportedOperationException e) {
```
massdosage, line 129: Why do we need this?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -56,13 +100,52 @@ public void removeTableParamsAndDrop(
...
     CaseInsensitiveMap caseInsensitiveParams = new CaseInsensitiveMap(tableParameters);
     return IS_EXTERNAL.equalsIgnoreCase((String) caseInsensitiveParams.get(EXTERNAL_KEY));
   }
+
+  private List<String> getPartitionLocations(CloseableMetaStoreClient client, Table table) {
+    List<String> locations = new ArrayList<>();
+    try {
+      PartitionIterator partitionIterator = new PartitionIterator(client, table, BATCH_SIZE);
+      while (partitionIterator.hasNext()) {
+        locations.add(partitionIterator.next().getSd().getLocation());
```
massdosage, line 144: Ah OK, so we do use the iterator but we then add them all to a list so this could still OOM. I think it's alright to do this for now but the "proper" way would be to use the iterator, not an intermediate collection.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -56,13 +100,52 @@ public void removeTableParamsAndDrop(
...
+      }
+    } catch (TException e) {
+      LOG.info("Could not list partitions for {}.{}.", table.getDbName(), table.getTableName());
```
massdosage, line 147: Log the exception.

Also, should we swallow all these exceptions with warnings or should we throw them and fail the replication?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulationClientFactoryManagerTest.java
```diff
@@ -0,0 +1,194 @@
...
+  private DataManipulationClientFactory clientFactory;
+  private Path sourceLocation;
+  private Path replicaLocation;
+  private final String s3Path = "s3://<path>";
+  private final String hdfsPath = "hdfs://<path>";
+  private final Map<String, Object> copierOptions = new HashMap<>();
+
+  @Before
+  public void setup() {
+    s3s3ClientFactory = new TestDataManipulationClientFactory();
```
massdosage, line 53: You can do this on line 39

massdosage: similar for lines 56, 59

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulationClientFactoryManagerTest.java
```diff
@@ -0,0 +1,194 @@
...
+    ((TestDataManipulationClientFactory) s3s3ClientFactory).setS3Client();
```
massdosage, line 54: Might be cleaner to have a small enum with the types S3_S3, S3_MAPREDUCE and HDFS and then have one method called something like `setManipulatorType`. You could also pass the type in to the constructor and then not need this extra method.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulationClientFactoryManagerTest.java
```diff
@@ -0,0 +1,194 @@
...
+        Arrays.asList(s3s3ClientFactory, s3MapReduceClientFactory, hdfsClientFactory));
+    sourceLocation = new Path(s3Path);
+  }
+
+  @Test
+  public void awsClientReturnedForS3S3Copy() {
+    replicaLocation = new Path(s3Path);
+    clientFactory = manager.getClientFactory(sourceLocation, replicaLocation, copierOptions);
+
+    Assert.assertTrue(((TestDataManipulationClientFactory) clientFactory).isS3S3Client());
```
massdosage, line 72: Static import `Assert.`

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulationClientFactoryManagerTest.java
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  @Test
+  public void awsMapReduceClientReturnedForHdfsS3Copy() {
+    sourceLocation = new Path(hdfsPath);
+    replicaLocation = new Path(s3Path);
+    clientFactory = manager.getClientFactory(sourceLocation, replicaLocation, copierOptions);
+
+    Assert.assertTrue(((TestDataManipulationClientFactory) clientFactory).isS3MapreduceClient());
```
massdosage, line 81: Is this checking anything other than the logic you've put in the test class itself? I don't think so?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableServiceTest.java
```diff
@@ -59,6 +72,10 @@ public void setUp() throws TException {
     table.setTableName(TABLE_NAME);
     table.setDbName(DB_NAME);
     when(client.getTable(DB_NAME, TABLE_NAME)).thenReturn(table);
+
+    storageDescriptor = new StorageDescriptor();
```
massdosage, line 76: If you're mocking this (line 64) then why are you doing this?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableServiceTest.java
```diff
@@ -132,4 +149,97 @@ public void removeParamsAndDropTableDoesNotExist() throws TException {
...
+    when(client.getPartitionsByNames(DB_NAME, TABLE_NAME, batch1)).thenReturn(partitionBatch1);
+    when(client.getPartitionsByNames(DB_NAME, TABLE_NAME, batch2)).thenReturn(Arrays.asList(partitionBatch2));
+    
+    table.setPartitionKeys(createFieldSchemaList(partitionNames));
+
+    service.dropTableAndData(client, DB_NAME, TABLE_NAME, dataManipulationClient);
+
+    verify(client).getTable(DB_NAME, TABLE_NAME);
+    verify(client).dropTable(DB_NAME, TABLE_NAME, false, true);
+    verify(dataManipulationClient).delete(PARTITION_LOCATION + "1");
```
massdosage, line 213: You could put this in a for loop from 1-11, would remove a few lines of code.

circus-train-distcp-copier/src/main/java/com/hotels/bdp/circustrain/distcpcopier/HdfsDataManipulationClient.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.hotels.bdp.circustrain.api.data.DataManipulationClient;
+
+public class HdfsDataManipulationClient implements DataManipulationClient {
+
+  private static final Logger LOG = LoggerFactory.getLogger(HdfsDataManipulationClient.class);
```
massdosage, line 30:
```suggestion
  private static final Logger log = LoggerFactory.getLogger(HdfsDataManipulationClient.class);
```

circus-train-distcp-copier/src/test/java/com/hotels/bdp/circustrain/distcpcopier/HdfsDataManipulationClientFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+  // This client factory will technically support all schemes including replicating to and from s3. However the s3
+  // client factories take higher precedence so this client factory wont be used for s3.
+  @Test
+  public void checkSupportsHdfs() {
+    sourceLocation = hdfsPath;
+    replicaLocation = hdfsPath;
+
+    boolean support = clientFactory.supportsSchemes(sourceLocation, replicaLocation);
+    assertTrue(support);
```
massdosage, line 53: You can do this in one line above.

circus-train-distcp-copier/src/test/java/com/hotels/bdp/circustrain/distcpcopier/HdfsDataManipulationClientFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Test
+  public void checkSupportsHdfsUpperCase() {
+    sourceLocation = hdfsPath.toUpperCase();
+    replicaLocation = hdfsPath.toUpperCase();
+
+    boolean support = clientFactory.supportsSchemes(sourceLocation, replicaLocation);
+    assertTrue(support);
```
massdosage, line 62: Ditto and below etc.

circus-train-s3-mapreduce-cp/src/test/java/com/hotels/bdp/circustrain/s3mapreducecp/aws/AwsMapreduceDataManipulationClientFactoryTest.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Test
+  public void checkSupportsHdfsToS3() {
+    sourceLocation = hdfsPath;
+    replicaLocation = s3Path;
+    sourceScheme = sourceLocation.toUri().getScheme();
+    replicaScheme = replicaLocation.toUri().getScheme();
+
+    boolean support = clientFactory.supportsSchemes(sourceScheme, replicaScheme);
```
massdosage, line 57: Same as in other test.

circus-train-s3-s3-copier/src/test/java/com/hotels/bdp/circustrain/s3s3copier/aws/AwsDataManipulationClientFactoryTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+  @Test
+  public void checkSupportsS3ToS3() {
+    sourceLocation = new Path(s3Path);
+    replicaLocation = new Path(s3Path);
+    sourceScheme = sourceLocation.toUri().getScheme();
+    replicaScheme = replicaLocation.toUri().getScheme();
+
+    boolean support = clientFactory.supportsSchemes(sourceScheme, replicaScheme);
+    assertTrue(support);
```
massdosage, line 59: etc.

README.md
JayGreeeen: Yeah that works for me. Think the name came from starting out with copying from s3->s3 and needing an s3client to "manipulate the data". Happy to change to this, it reads nicer.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
JayGreeeen: Its technically no different, but it gets the table object and passes that into the `dropTable` method. I added the `dropTable` method and made it take in a table object because otherwise when `dropTableAndData` is called it would get the table object, delete partition info, then call `dropTable` which will get the table object again and then drop it. So by doing it this way it reduces the number of times that `client.getTable` is called.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
JayGreeeen: Do you mean the hcommon PartitionIterator trick? b/c if so then it does use that in the `getPartitionLocations` method at the bottom of the class.
`PartitionIterator partitionIterator = new PartitionIterator(client, table, BATCH_SIZE);`

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
JayGreeeen: I don't mind. It technically wont affect the replication, but you could end up with a lot of orphaned data which wont be cleaned up.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulationClientFactoryManagerTest.java
JayGreeeen: Not really, just checking that it will go through the list of factories, check if the replication is supported and returns the right one. And that it returns the one specified in the copier options (if any).

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
patduin: should't we just throw the exception and stop?
We haven't deleted the existing one so writing again will probably fail anyway.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
patduin: let's fix this please, we've hit this in the past, it's better to make this resistant now since we're testing it actively and actually have it fresh in our heads.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
patduin: I think all exceptions in this class except ones we can recover from should be thrown and stop the replication. Logging them serves little purpose, no one is going to look through the log if the end replication reports OK. If the purpose is alerting the user to manually clean up, best to just fail. If this becomes unfeasible e.g. lots of failures we can build retry mechanics etc.. but out of scope now.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
patduin: Yes this is not useful to log and might even be confusing.

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
patduin: I think we should log on line 124 before doing the delete which location we're trying to delete and then let the exception bubble up. Why try all of them if 1 fails and 99 succeeds shouldn't we stop the replication?

massdosage: OK, as long as the approach is consistent, I'm fine with it being "as soon as anything goes wrong, ABORT AND FAIL!"

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/data/DataManipulator.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.api.data;
+
+import java.io.IOException;
+
+public interface DataManipulator {
+
+  /**
+   * Returns boolean stating whether the deletion was successful or not. May through an error while attempting to delete
```
massdosage, line 23: Please document the "may throw an error" part using the standard Javadoc for exceptions thrown.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/AwsDataManipulator.java
```diff
@@ -0,0 +1,83 @@
...
+import com.amazonaws.services.s3.AmazonS3URI;
+import com.amazonaws.services.s3.model.DeleteObjectsRequest;
+import com.amazonaws.services.s3.model.DeleteObjectsRequest.KeyVersion;
+import com.amazonaws.services.s3.model.DeleteObjectsResult;
+import com.amazonaws.services.s3.model.ObjectListing;
+import com.amazonaws.services.s3.model.S3ObjectSummary;
+
+import com.hotels.bdp.circustrain.api.data.DataManipulator;
+
+public class AwsDataManipulator implements DataManipulator {
```
massdosage, line 38:
```suggestion
public class S3DataManipulator implements DataManipulator {
```
feels like a more accurate/specific name to me

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/data/DataManipulatorFactoryManager.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package com.hotels.bdp.circustrain.api.data;
+
+import java.util.Map;
+
+import org.apache.hadoop.fs.Path;
+
+public interface DataManipulatorFactoryManager {
+
+  public DataManipulatorFactory getClientFactory(
```
massdosage, line 24:
```suggestion
  public DataManipulatorFactory getFactory(
```
(since we've removed the word "client" elsewhere)

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManager.java
```diff
@@ -0,0 +1,89 @@
...
+      Path replicaTableLocation,
+      Map<String, Object> copierOptions) {
+    String replicaLocation = replicaTableLocation.toUri().getScheme();
+    String sourceLocation = sourceTableLocation.toUri().getScheme();
+
+    if (copierOptions.containsKey(DATA_MANIPULATOR_FACTORY_CLASS)) {
+      for (DataManipulatorFactory clientFactory : dataManipulatorFactories) {
+        final String clientFactoryClassName = clientFactory.getClass().getName();
+        if (clientFactoryClassName.equals(copierOptions.get(DATA_MANIPULATOR_FACTORY_CLASS).toString())) {
+          log.debug("Found ClientFactory '{}' using config", clientFactoryClassName);
```
massdosage, line 71:
```suggestion
          log.debug("Found DataManipulatorFactory '{}' using config", clientFactoryClassName);
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManager.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Autowired
+  public DefaultDataManipulatorFactoryManager(List<DataManipulatorFactory> factories) {
+    this.dataManipulatorFactories = factories;
+  }
+
+  @PostConstruct
+  void postConstruct() {
+    log.debug("Initialized with {} DataManipulatorFactories", dataManipulatorFactories.size());
+    for (DataManipulatorFactory clientFactory : dataManipulatorFactories) {
```
massdosage, line 54:
```suggestion
    for (DataManipulatorFactory factory : dataManipulatorFactories) {
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManager.java
```diff
@@ -0,0 +1,89 @@
...
+  @Override
+  public DataManipulatorFactory getClientFactory(
+      Path sourceTableLocation,
+      Path replicaTableLocation,
+      Map<String, Object> copierOptions) {
+    String replicaLocation = replicaTableLocation.toUri().getScheme();
+    String sourceLocation = sourceTableLocation.toUri().getScheme();
+
+    if (copierOptions.containsKey(DATA_MANIPULATOR_FACTORY_CLASS)) {
+      for (DataManipulatorFactory clientFactory : dataManipulatorFactories) {
```
massdosage, line 68:
```suggestion
      for (DataManipulatorFactory factory : dataManipulatorFactories) {
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManager.java
```diff
@@ -0,0 +1,89 @@
...
+        final String clientFactoryClassName = clientFactory.getClass().getName();
```
massdosage, line 69:
```suggestion
        final String factoryClassName = clientFactory.getClass().getName();
```

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -435,6 +427,9 @@ public ReplicaLocationManager getLocationManager(
       SourceLocationManager sourceLocationManager) {
     CleanupLocationManager cleanupLocationManager = CleanupLocationManagerFactory
         .newInstance(eventId, housekeepingListener, replicaCatalogListener, tableReplication);
+
+    sourceLocationManager.getTableLocation();
```
massdosage, line 431: What does this line do? The result isn't being assigned to anything - is it needed?

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/hive/DropTableService.java
```diff
@@ -25,28 +26,62 @@
...
   private static final String IS_EXTERNAL = "TRUE";
+  private static final short BATCH_SIZE = (short) 1000;
 
   /**
    * Removes all parameters from a table before dropping the table.
+   * 
+   * @throws Exception if the table can't be deleted.
    */
-  public void removeTableParamsAndDrop(
+  public void dropTable(CloseableMetaStoreClient client, String databaseName, String tableName) throws Exception {
```
massdosage, line 45: I think this and the changes below are much clearer now, also better using the partition iterator properly, nice.

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+import org.mockito.runners.MockitoJUnitRunner;
+
+import com.hotels.bdp.circustrain.api.data.DataManipulator;
+import com.hotels.bdp.circustrain.api.data.DataManipulatorFactory;
+import com.hotels.bdp.circustrain.api.data.DataManipulatorFactoryManager;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DefaultDataManipulatorFactoryManagerTest {
+
+  private DataManipulatorFactory s3s3DataManipulatorFactory = new TestDataManipulatorFactory(S3_S3);;
```
massdosage, line 44: Woah, what's up with those double semi-colons?!?!? Crrraazzyy, they need to GO! (and below)

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+  @Before
+  public void setup() {
+    manager = new DefaultDataManipulatorFactoryManager(
+        Arrays.asList(s3s3DataManipulatorFactory, s3MapReduceDataManipulatorFactory, hdfsDataManipulatorFactory));
+    sourceLocation = new Path(s3Path);
+  }
+
+  @Test
+  public void awsClientReturnedForS3S3Copy() {
```
massdosage, line 64:
```suggestion
  public void s3ManipulatorReturnedForS3S3Copy() {
```
(and similar renames for methods below since we don't use the word "client" for these anymore)

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    replicaLocation = new Path(s3Path);
+    clientFactory = manager.getClientFactory(sourceLocation, replicaLocation, copierOptions);
+
+    assertTrue(((TestDataManipulatorFactory) clientFactory).getType() == S3_S3);
```
massdosage, line 68: I'd prefer using `assertEquals` here, you get a better error message from JUnit if this fails rather than just "wasn't true".

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  @Test
+  public void awsMapReduceClientReturnedForHdfsS3Copy() {
+    sourceLocation = new Path(hdfsPath);
+    replicaLocation = new Path(s3Path);
+    clientFactory = manager.getClientFactory(sourceLocation, replicaLocation, copierOptions);
+
+    assertTrue(((TestDataManipulatorFactory) clientFactory).getType() == S3_MAPREDUCE);
```
massdosage, line 77: Ditto (and etc. below)

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+  @Test(expected = UnsupportedOperationException.class)
+  public void noSupportingFactory() {
+    replicaLocation = new Path("<path>");
+    clientFactory = manager.getClientFactory(sourceLocation, replicaLocation, copierOptions);
+
+    assertTrue(((TestDataManipulatorFactory) clientFactory).getType() == HDFS);
+  }
+
+  public enum DataManipulatorType {
```
massdosage, line 109: Does this need to be public?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    S3_S3,
+    S3_MAPREDUCE,
+    HDFS
+  }
+
+  class TestDataManipulatorFactory implements DataManipulatorFactory {
```
massdosage, line 115: Can this be private?

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
```diff
@@ -187,7 +192,7 @@ private Replica newReplica(TableReplication tableReplication) {
   }
 
   @Test
-  public void alteringExistingUnpartitionedReplicaTableSucceeds() throws TException, IOException {
+  public void alteringExistingUnpartitionedReplicaTableSucceeds() throws Exception, IOException {
```
massdosage, line 195:
```suggestion
  public void alteringExistingUnpartitionedReplicaTableSucceeds() throws Exception {
```

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/replica/ReplicaTest.java
```diff
@@ -196,7 +201,7 @@ public void alteringExistingUnpartitionedReplicaTableSucceeds() throws TExceptio
   }
 
   @Test
-  public void alteringExistingUnpartitionedReplicaTableWithNoStatsSucceeds() throws TException, IOException {
+  public void alteringExistingUnpartitionedReplicaTableWithNoStatsSucceeds() throws Exception, IOException {
```
massdosage, line 204:
```suggestion
  public void alteringExistingUnpartitionedReplicaTableWithNoStatsSucceeds() throws Exception {
```
etc. below too

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainReplicationModeIntegrationTest.java
```diff
@@ -230,18 +242,26 @@ public void partitionedTableFullOverwrite() throws Exception {
...
-            .listPartitions(DATABASE, TARGET_PARTITIONED_TABLE, (short) -1);
-        assertThat(partitions.size(), is(2));
-        assertThat(partitions.get(0).getValues(), is(Arrays.asList("Asia", "China")));
-        assertThat(partitions.get(1).getValues(), is(Arrays.asList("Europe", "UK")));
+        // Assert files have been copied from source and replica file is deleted
+        List<S3ObjectSummary> replicaFiles = TestUtils.listObjects(s3Client, "replica");
+        assertThat(replicaFiles.size(), is(2));
+
+        String partitionAsia = "continent=Asia/country=China";
+        String asiafileKey = String
```
massdosage, line 257:
```suggestion
        String asiaFileKey = String
```

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainReplicationModeIntegrationTest.java
```diff
@@ -230,18 +242,26 @@ public void partitionedTableFullOverwrite() throws Exception {
...
+            .format("%s/%s/%s/%s/%s", DATABASE, TARGET_PARTITIONED_TABLE, eventId, partitionAsia, PART_00000);
+        assertThat(replicaFiles.get(0).getKey(), is(asiafileKey));
+
+        String partitionEurope = "continent=Europe/country=UK";
+        String europefileKey = String
```
massdosage, line 262:
```suggestion
        String europeFileKey = String
```

circus-train-s3-mapreduce-cp/src/main/java/com/hotels/bdp/circustrain/s3mapreducecp/aws/AwsMapreduceDataManipulatorFactory.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  // The hdfs -> s3 data manipulator doesn't need to use the path for the client.
+  @Override
+  public DataManipulator newInstance(Path path, Map<String, Object> copierOptions) {
+    return new AwsDataManipulator(s3ClientFactory.newInstance(conf));
+  }
+
+  /**
+   * Supports copying from hdfs/ a file to s3.
```
massdosage, line 57:
```suggestion
   * Supports copying from HDFS to S3.
```

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/data/DataManipulatorFactoryManager.java
JayGreeeen: oh dammit, I missed some! Nice spot!

massdosage: > oh dammit, I missed some! Nice spot!

I try my best!

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManagerTest.java
JayGreeeen: 😱 copy-paste fail!!!

circus-train-core/src/test/java/com/hotels/bdp/circustrain/core/data/DefaultDataManipulatorFactoryManagerTest.java
JayGreeeen: Looks like assertEquals is deprecated!

massdosage: Are you sure? We use it all over the show, what does it suggest you use instead in the deprecation warning?

JayGreeeen: `The method assertEquals(Object, Object) from the type Assert is deprecated. Add @SuppressWarnings 'deprecation' to 's3ManipulatorReturnedForS3S3Copy()'`

massdosage: `assertEquals(S3_S3, ((TestDataManipulatorFactory) clientFactory).getType());` looks OK to me? I don't see a warning in Eclipse?

JayGreeeen: hmm yeah that works for me too. It's when I do `Assert.assertEquals..` it has an issue. Weird. Will change them 👍🏻

massdosage: Which `Assert` class do you import when you do that? My guess is you're by mistake pulling in an old Junit Assert class where everything is deprecated.

JayGreeeen: ahh yeah looks like it, the one with everything deprecated is inside `junit.framework` but the Assert class that works is in `org.junit`.

massdosage:
Fixes the "data" part of #183

PR #188, opened 2020-06-15
Schema evolution tests

max-jacobs:
### :link: Related Issues


pom.xml
```diff
@@ -54,7 +54,7 @@
     <httpcomponents.httpclient.version>4.5.11</httpcomponents.httpclient.version>
     <jackson.version>2.10.0</jackson.version>
     <!-- END: AWS jdk version + dependencies -->
-    <hadoop.version>2.7.1</hadoop.version>
+    <hadoop.version>2.8.1</hadoop.version>
```
massdosage, line 57: Hive 2.3.7 supposedly depends on Hadoop 2.7.2 so why did you need to move to 2.8.1?

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/utils/ThriftMetastoreServerRuleExtension.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.integration.utils;
+
+import org.apache.hadoop.hive.conf.HiveConf;
+import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
+
+import com.hotels.beeju.ThriftHiveMetaStoreJUnitRule;
+
+public class ThriftMetastoreServerRuleExtension extends ThriftHiveMetaStoreJUnitRule {
```
massdosage, line 23: Is this class just a convenience to override the conf values?

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainParquetSchemaEvolutionIntegrationTest.java
```diff
@@ -0,0 +1,682 @@
+/**
```
massdosage, line 1: So is this the class you're not so sure about adding as it increases the test running time so much? One option could be to use a Junit 4 Category (`SchemaTests`) and then turn off this category by default when Maven runs the test. Then in the SCHEMA_EVOLUTION.md you could add a section about the tests and show how one can run them on the CLI by telling Maven to specifically run tests with that category. I assume that's the idea here - they're a safety net that we should never need but they could be useful every now and then just to make sure we haven't broken any of the schema evolution rules as we understand them today?

pom.xml
max-jacobs: Hmm yep I think i did this to fix some Parquet writer error, will revert as this was fixed elsewhere

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/utils/ThriftMetastoreServerRuleExtension.java
max-jacobs: Pretty much yep, `beforeTest()` is called from within beeju, so this is where we will initialise beeju with the hiverunner metastore

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainParquetSchemaEvolutionIntegrationTest.java
max-jacobs: Yep this is the one. In total these tests take ~10 mins on my machine. Plus they all pass (when in reality the non supported schema evolutions shouldn't - for those cases I've made the tests expect an exception to be thrown).

They're a safety net, and also useful if a feature is ever added to CT to support some of the currently failing evolutions. They're essentially supporting tests for the added readme as "proof".

max-jacobs: I've added junit category as suggested, added to the doc too. Let me know what you think, thanks!

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/utils/ThriftMetastoreServerRuleExtension.java
massdosage: OK

circus-train-common-test/src/main/java/com/hotels/bdp/circustrain/common/test/base/CircusTrainRunner.java
```diff
@@ -146,6 +160,7 @@ private CircusTrainRunner(
       String graphiteUri,
       File housekeepingDbLocation) {
     this.databaseName = databaseName;
+    this.replicaDatabaseName = replicaDatabaseName;
```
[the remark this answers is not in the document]
max-jacobs: Yep its used in the new integration tests. It seemed easier with Hiverunner to just setup one metastore with a source db and a replica db. So I had to change the runner here to be able to set different dbs. The other integration tests don't use this, so the `replicaDatabaseName` just defaults to the `databaseName` if empty.

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainParquetSchemaEvolutionIntegrationTest.java
```diff
@@ -0,0 +1,685 @@
...
+    File config = dataFolder.getFile("partitioned-single-table-one-partition.yml");
+    runner.run(config.getAbsolutePath());
+  }
+
+  private void runDataChecks(Schema schema, List<String> expectedData) throws Exception {
+    try {
+      assertTable(thriftMetaStoreRule.newClient(), schema, SOURCE_DB, TABLE, expectedData);
+      assertTable(thriftMetaStoreRule.newClient(), schema, REPLICA_DB, TABLE, expectedData);
+    } catch (Exception e) {
+      throw new Exception("Test failed", e);
```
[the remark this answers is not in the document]
max-jacobs: Good point, no benefit, will remove

circus-train-integration-tests/pom.xml
```diff
@@ -198,6 +264,7 @@
         <configuration>
           <forkCount>1</forkCount>
           <reuseForks>false</reuseForks>
+          <groups>${groups}</groups>
```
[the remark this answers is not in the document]
max-jacobs: Yeah thats nicer, will update

circus-train-common-test/src/main/java/com/hotels/bdp/circustrain/common/test/base/CircusTrainRunner.java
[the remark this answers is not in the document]
massdosage: Ah, OK.

circus-train-integration-tests/pom.xml
```diff
@@ -14,9 +15,20 @@
...
   <dependencies>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-api</artifactId>
+      <version>2.5</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+      <version>2.5</version>
```
massdosage, line 30: Minor, but since you have this twice and they're related you could make this a property.

circus-train-integration-tests/pom.xml
```diff
@@ -188,6 +206,39 @@
...
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <version>3.1.0</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.esotericsoftware</groupId>
+      <artifactId>kryo</artifactId>
+      <version>4.0.2</version>
```
massdosage, line 224: Should this be test or provided scope? I'd be concerned of possible classpath clash if this ends up in CT where it didn't beore.

max-jacobs: will move this to the test scope

circus-train-integration-tests/pom.xml
```diff
@@ -135,12 +148,18 @@
...
+          <artifactId>javax.servlet-api</artifactId>
+        </exclusion>
+      </exclusions>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.5</version>
+      <version>${log4j-api.version}</version>
```
massdosage, line 162: I'm pretty sure you didn't mean to do this!

max-jacobs: :(

circus-train-integration-tests/pom.xml
```diff
@@ -14,9 +15,21 @@
     <hadoop.version>2.8.1</hadoop.version>
     <!-- override for hadoop-common dependency that is used in this test -->
     <guava.version>17.0</guava.version>
+    <log4j-api.version>2.5</log4j-api.version>
```
massdosage, line 18:
```suggestion
    <log4j.version>2.5</log4j.version>
```
(it's not just the api)

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/utils/ThriftMetastoreServerRuleExtension.java
patduin: We should push this to Beeju once and fix it there (We also do this in other internal projects).

massdosage: Agreed, can someone do that now? It not can we add an issue over at BeeJU and refer to this?

PR #193, opened 2020-07-01
Prep for release

JayGreeeen:
### :link: Related Issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [16.2.0] - 2020-06-30
+## [16.2.0] - 2020-07-01
 ### Changed
 * Changed version of `hive.version` to `2.3.7` (was `2.3.2`). This allows Circus Train to be used on JDK>=9.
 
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## [16.2.0] - 2020-06-30
+## [16.2.0] - 2020-07-01
```
abhimanyugupta07, line 1: So, you are releasing 16.2.0 but snapshot is 16.1.1 ?

circus-train-api/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>com.hotels</groupId>
     <artifactId>circus-train-parent</artifactId>
-    <version>16.2.1-SNAPSHOT</version>
+    <version>16.1.1-SNAPSHOT</version>
```
massdosage, line 7: Shouldn't this be 16.2.0-SNAPSHOT? (although you can override it in Jenkins when you do the release so doesn't matter too much)

JayGreeeen: I just reverted the changes that the maven release made and put it back to what it was before, so when I release I'll add the version then.

CHANGELOG.md
JayGreeeen: Yeah that will be changed when I perform the release

PR #196, opened 2020-08-21
Copier context

### LINKED ISSUE #195: Make more configuration values available to Copiers
**Is your feature request related to a problem? Please describe.**
When developing a downstream Copier for Circus Train we discovered that we require the Hive database and table name in order for the copier to be able to retrieve the Table's tags. Currently the copiers are only passed information about the paths which isn't enough for our copier.

**Describe the solution you'd like**
Ideally we would have more of the table replication context available to the copier but the immediate need is these two fields.

**Describe alternatives you've considered**
The only way that we can think of currently to get these values is to pass them in as custom copier options but this then means that they are duplicated in the config which isn't ideal.

massdosage:
Fixes #195 

This PR introduces a `CopierContext` class to hold the values that were being passed to the `CopierFactory` before, and then adds a `TableReplication` to this which in turns holds the additional values that were requested in the issue (source database and table names). This allows us to add more values to CopierContext in the past without breaking backwards compatibility which is the situation we're currently in due to the interface methods directly containing the values as arguments.

The changes in this PR are currently backwards-compatible by using a default method in the `CopierFactory` interface but I'll add some comments on this as another option would be to make a backwards *incompatible* change as it's quite easy to resolve downstream.

### DIFF
--- a/circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
+++ b/circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
@@ -15,6 +15,7 @@
  */
 package com.hotels.bdp.circustrain.api.copier;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -26,28 +27,35 @@ public class CopierContext {
 
   private String eventId;
   private Path sourceBaseLocation;
-  private List<Path> sourceSubLocations;
+  private List<Path> sourceSubLocations = Collections.<Path>emptyList();
   private Path replicaLocation;
   private Map<String, Object> copierOptions;
   private TableReplication tableReplication;
-  private String sourceDatabaseName;
-  private String sourceTableName;
-  
-  //TODO: below constructor added to make it easier to construct these based on how we used to do it, probably replace with a builder instead.
-  public CopierContext(String eventId, Path sourceBaseLocation, List<Path> sourceSubLocations, Path replicaLocation, Map<String, Object> copierOptions) {
+
+  public CopierContext(
+      String eventId,
+      Path sourceBaseLocation,
+      List<Path> sourceSubLocations,
+      Path replicaLocation,
+      Map<String, Object> copierOptions) {
     this.eventId = eventId;
     this.sourceBaseLocation = sourceBaseLocation;
     this.sourceSubLocations = sourceSubLocations;
     this.replicaLocation = replicaLocation;
     this.copierOptions = copierOptions;
   }
-  
-  public CopierContext(String eventId, Path sourceBaseLocation, Path replicaLocation, Map<String, Object> copierOptions) {
+
+  public CopierContext(
+      String eventId,
+      Path sourceBaseLocation,
+      Path replicaLocation,
+      Map<String, Object> copierOptions) {
     this.eventId = eventId;
     this.sourceBaseLocation = sourceBaseLocation;
     this.replicaLocation = replicaLocation;
     this.copierOptions = copierOptions;
   }
+
   public String getEventId() {
     return eventId;
   }
@@ -88,18 +96,12 @@ public class CopierContext {
     this.copierOptions = copierOptions;
   }
 
-  //TODO: decide on this 
   public void setTableReplication(TableReplication tableReplication) {
     this.tableReplication = tableReplication;
   }
-  //OR
-  public void setSourceDatabaseName(String sourceDatabaseName) {
-    this.sourceDatabaseName = sourceDatabaseName;
-    
+
+  public TableReplication getTableReplication() {
+    return tableReplication;
   }
-  public void setSourceTableName(String sourceTableName) {
-    this.sourceTableName = sourceTableName;    
-  }
-  
 
 }
--- a/circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
+++ b/circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
@@ -1,0 +1,98 @@
+/**
+ * Copyright (C) 2016-2020 Expedia, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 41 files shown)


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/ReplicationFactoryImpl.java
```diff
@@ -134,29 +124,25 @@ private Replication createPartitionedTableReplication(
   private Replication createUnpartitionedTableReplication(
       TableReplication tableReplication,
       Source source,
-      Replica replica,
-      String sourceDatabaseName,
-      String sourceTableName,
-      String replicaDatabaseName,
-      String replicaTableName,
-      String replicaTableLocation) {
+      Replica replica) {
```
nvitucci, line 127: Nice simplification.

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierFactory.java
```diff
@@ -25,13 +25,28 @@
...
+   */
+  default Copier newInstance(CopierContext copierContext) {
+    //TODO: this is only here for backwards compatibility with CopierFactorys using older versions of Circus Train, when the below
+    //deprecated methods are removed so should this default implementation
+    return newInstance(copierContext.getEventId(), copierContext.getSourceBaseLocation(), copierContext.getSourceSubLocations(), copierContext.getReplicaLocation(), 
+        copierContext.getCopierOptions());
+  }
+
+  /**
+   * @deprecated As of release 16.3.0, replaced by {@link #newInstance(CopierContext)}.
```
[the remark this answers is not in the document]
nvitucci: I'd still rather deprecate it and, in case, have it removed later on.

nvitucci (approved):
It looks good to me. The `TableReplication` simplifies the code further. 
I would still deprecate the "old" method rather then removing it, and possibly remove it in a later release.

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierFactory.java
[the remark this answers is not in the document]
patduin: deprecate and remove is the more friendly way, I'm ok either way though it's not a big change to do.

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.hadoop.fs.Path;
+
+import com.hotels.bdp.circustrain.api.conf.TableReplication;
+
+public class CopierContext {
```
patduin, line 26: thoughts on making this immutable? Is there a risk of a copier accidentally change one of these fields which might affect further copiers in the chain?

massdosage: Yeah, I thought about making this immutable and having a builder instead of this big constructor but this was quicker and easier to do in terms of updating the existing code. However the TableReplication isn't immutable so we'd only be halfway there unless I looked into that too (not sure that's really possible as it maps directly to the config which needs to set values).

If we didn't want to use a builder I could add 2 more constructors which allow the TableReplication to be passed in too and then I make immutable copies of the List and Map member variables and we live with TableReplication not being immutable.

I can do either (or if someone has better suggestions) if we think it's worth it?

patduin: I think since this is a new one we make the good example and make it immutable. I'm ok leaving TableReplication for another time as that will be backward incompatible change I guess.

nvitucci: Isn't there any place where the context may actually need to be updated?

patduin: yeah I was also thinking about that, not sure what we want to achieve. The order in which copiers are run is guaranteed by spring so you could have a copier relying on another copier updating the context. If that's they way we want to go then obviously we don't want to make it immutable. Could argue either way I guess :)

massdosage: At the moment I don't see any of the copiers updating the context but yes, you could argue that that could be a useful feature ;) My suggestion would be to follow our usual best practice and make them immutable for now, if we ever find we need some way of modifying/sharing context between copiers we could break that and make them mutable (or at that point perhaps some other solution presents itself).

nvitucci: If it keeps on being backward-compatible and not too difficult to do, go ahead with making it immutable then. All the other things being equal, immutable objects are always preferable.

massdosage: I've changed it to be as immutable as I think I can :)

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierFactory.java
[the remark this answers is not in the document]
massdosage: OK, let's go with a 16.3.0 release that is backwards compatible, I'll update the changelog accordingly.

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
nvitucci: Looks good!

patduin: the class needs to be final

massdosage: It does! Good spot!

patduin: and because TableReplication isn't immutable you need to copy it as well. Path also isn't :/ it's a pandemic. Not sure how much you want to open that can of worms. I can live with a not fully immutable class for the sake of pragmatism.

massdosage: Yeah, you're right, I think we either leave as it is now and live with a decent _attempt_ but `TableReplication` and `Path` are mutable or we just roll the whole set of immutable changes back.

patduin: I'm ok with this.

patduin (approved):

PR #198, opened 2020-09-07
Added source table and partitions to CopierContext

nvitucci:
Adding `Table sourceTable` and `List<Partition> sourcePartitions` to `CopierContext` in order to propagate table and partition filter information down to the copiers.

### DIFF
--- a/circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
+++ b/circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
@@ -20,6 +20,8 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.metastore.api.Partition;
+import org.apache.hadoop.hive.metastore.api.Table;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -34,6 +36,8 @@ public final class CopierContext {
   private Path replicaLocation;
   private Map<String, Object> copierOptions;
   private TableReplication tableReplication;
+  private Table sourceTable;
+  private List<Partition> sourcePartitions;
 
   public CopierContext(
       TableReplication tableReplication,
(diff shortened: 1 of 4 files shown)


massdosage (commented):
Code changes look good, needs unit tests and CHANGELOG. Let's try this out downstream in the project which requires this before merging though, in case there are other changes we need too.

circus-train-api/src/main/java/com/hotels/bdp/circustrain/api/copier/CopierContext.java
```diff
@@ -50,6 +56,40 @@ public CopierContext(
     }
     this.replicaLocation = replicaLocation;
     this.copierOptions = ImmutableMap.copyOf(copierOptions);
+    this.sourceTable = sourceTable;
+    this.sourcePartitions = sourcePartitions;
+  }
+
+  public CopierContext(
```
patduin, line 63: Can we just remove these three constructors and stick with changing this one. It's very confusing what you need to actually make. We're not using

circus-train-api/src/test/java/com/hotels/bdp/circustrain/api/copier/CopierContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Before
+  public void init() {
+    copierContext = new CopierContext(tableReplication, eventId, sourceLocation, sourceSubLocations,
+        replicaLocation, copierOptions, sourceTable, sourcePartitions);
+  }
+
+  @Test
+  public void getters() {
+    assertThat(copierContext.getEventId(), is("1"));
```
massdosage, line 59:
```suggestion
    assertThat(copierContext.getEventId(), is(eventId));
```
^ maybe better to re-use the actual objects for the assertions? That way if we ever change them we only need to change them in on place instead of two.

circus-train-api/src/test/java/com/hotels/bdp/circustrain/api/copier/CopierContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import com.hotels.bdp.circustrain.api.conf.TableReplication;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CopierContextTest {
+
+  private @Mock TableReplication tableReplication;
+  private @Mock Map<String, Object> copierOptions;
```
massdosage, line 40:
```suggestion
  private Map<String, Object> copierOptions = new HashMap();
```
Really minor, but since we're not actually using any of the mock functionality and it's a simple one liner, might as well use a real object here.

massdosage (changes requested):
Looks good, a couple of really minor comments from me. Then change this to a non-WIP PR and update the CHANGELOG and then I think this is ready to go in.

massdosage (approved):

PR #208, opened 2020-12-08
Add check for empty list delete

andreeapad:


circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/S3DataManipulator.java
```diff
@@ -54,6 +54,11 @@ public boolean delete(String path) {
     List<KeyVersion> keysToDelete = getKeysToDelete(bucket, uri.getKey());
     log.debug("Deleting keys: {}", keysToDelete.stream().map(k -> k.getKey()).collect(Collectors.toList()));
 
+    if (keysToDelete.isEmpty()) {
+      log.info("Nothing to delete at location: {}", path);
+      return false;
```
nvitucci, line 59: Is `false` the most appropriate return value here? Although this is a special case, we have that 0 files have been deleted out of 0 requested (unless the return value is also an indication that a `delete` operation did actually take place).

andreeapad: Hmm, I'm not actually sure, I couldn't find the result of this method used anywhere

nvitucci: Same here, the only call I have found for this method discards the return value. @patduin, what do you think?

nvitucci (approved):
Looks good. My only doubt is on the return value, but given that it is not used anywhere it can be changed later.

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/S3DataManipulator.java
patduin: I think it's ok, false is what `io.File.delete()` does.

Unfortunately this class does not have a test, luckily that will soon be fixed right @andreeapad ?!? :)

andreeapad: yes, good point @patduin  😄

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
```diff
@@ -0,0 +1,88 @@
...
+        .build();
+    return newClient;
+  }
+
+  @Test
+  public void deleteFails() {
+    s3Client.createBucket(EMPTY_BUCKET);
+    boolean result = s3DataManipulator.delete(EMPTY_PATH);
+    assertThat(result, is(false));
+  }
```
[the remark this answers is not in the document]
andreeapad: yes, I updated the tests 👍

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/S3DataManipulator.java
```diff
@@ -54,6 +54,11 @@ public boolean delete(String path) {
     List<KeyVersion> keysToDelete = getKeysToDelete(bucket, uri.getKey());
     log.debug("Deleting keys: {}", keysToDelete.stream().map(k -> k.getKey()).collect(Collectors.toList()));
```
massdosage, line 55: Maybe move this to line 61 as there's no point outputting this if there's nothing to delete?

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
```diff
@@ -0,0 +1,100 @@
...
+  public void deleteFolderSucceeds() throws IOException {
+    s3Client.createBucket(BUCKET);
+    File inputData = temp.newFile("data");
+    s3Client.putObject(BUCKET, FOLDER, inputData);
+    boolean result = s3DataManipulator.delete(BUCKET_PATH + "/" + FOLDER);
+    assertThat(result, is(true));
+  }
+
+  @Test
+  public void deleteInexistentFolderFails() {
```
massdosage, line 81:
```suggestion
  public void deleteNonexistentFolderFails() {
```

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
```diff
@@ -0,0 +1,100 @@
...
+    s3Client.createBucket(BUCKET);
+    boolean result = s3DataManipulator.delete(BUCKET_PATH + "/inexistent-folder");
```
massdosage, line 83:
```suggestion
    boolean result = s3DataManipulator.delete(BUCKET_PATH + "/nonexistent-folder");
```

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
```diff
@@ -0,0 +1,100 @@
...
+    assertThat(result, is(false));
+  }
+
+  @Test
+  public void deleteBucketSucceeds() {
```
massdosage, line 88: Method says "succeeds" but assertion is on method returning `false`. Is the behaviour here that it will return false as it can't actually delete a bucket?

massdosage (changes requested):

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
andreeapad: oops, it should be `fails` in this case, I have to modify the test to make it succeed actually

circus-train-aws/src/main/java/com/hotels/bdp/circustrain/aws/S3DataManipulator.java
andreeapad: yeah

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+@RunWith(MockitoJUnitRunner.class)
+public class S3DataManipulatorTest {
+
+  private static final String AWS_ACCESS_KEY = "access";
+  private static final String AWS_SECRET_KEY = "secret";
+  private static final String BUCKET = "bucket";
+  private static final String BUCKET_PATH = "s3://" + BUCKET;
+  private static final String FOLDER = "folder";
+  private static final String EMPTY_BUCKET = "empty-bucket";
```
massdosage, line 46: I think this is actually non-existent bucket not empty right? As we don't create it anywhere.

massdosage (changes requested):

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
andreeapad: We create it in the last test. If we try to delete a non-existent bucket, we get an AmazonS3Exception. Should I add a test for that as well?

massdosage: Ah you're right, sorry,  I missed that. Might as well add a test for a non-existent bucket, it's an edge case we should never run into but to be thorough would be good to test it.

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
```diff
@@ -0,0 +1,105 @@
...
+  @Test
+  public void deleteEmptyBucketFails() {
+    s3Client.createBucket(EMPTY_BUCKET);
+    boolean result = s3DataManipulator.delete("s3://" + EMPTY_BUCKET);
+    assertThat(result, is(false));
+  }
+
+  @Test(expected = AmazonS3Exception.class)
+  public void deleteNonExistentBucketThrowsException() {
+    boolean result = s3DataManipulator.delete("s3://" + "nonexistent-bucket");
```
massdosage, line 102: If this throws an exception you don't need to store the result or assert on it as that code will never execute.

andreeapad: true, copy pasted and forgot to update 😄

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
```diff
@@ -0,0 +1,104 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hotels.bdp.circustrain.aws;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.*;
```
massdosage, line 19: OMG, not sure how this slipped through but we don't want `*` imports, please get your IDE to expand these.

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
```diff
@@ -0,0 +1,104 @@
...
+  @Test
+  public void deleteEmptyBucketFails() {
+    s3Client.createBucket(EMPTY_BUCKET);
+    boolean result = s3DataManipulator.delete("s3://" + EMPTY_BUCKET);
+    assertThat(result, is(false));
+  }
+
+  @Test(expected = AmazonS3Exception.class)
+  public void deleteNonExistentBucketThrowsException() {
+    s3DataManipulator.delete("s3://" + "nonexistent-bucket");
```
massdosage, line 102: You don't need the `+` :)

andreeapad: ahhhhh

circus-train-aws/src/test/java/com/hotels/bdp/circustrain/aws/S3DataManipulatorTest.java
andreeapad: fixed, not sure what happened either

massdosage (approved):

patduin (approved):

PR #211, opened 2021-02-12
Get build working on Java 11

massdosage:
This PR covers a number of changes required to get Circus Train building successfully using Java 11, as well as some cleanup and improvements, here's a summary of the changes:

* Add explicit imports on libraries no longer present in Java 11 (`javax.annotation-api`, `jaxb-api`).
* Exclude dependencies on `jdk.tools` as this is no longer present in Java 11.
* Removed some usages of `powermock` which didn't play nicely with Java 11.
* Upgrade to latest 2.x versions of Mockito and PowerMock for Java 11 support.
* Proactively move to non-deprecated code in JUnit, Mockito and Hamcrest (brought to you by `sed` and `grep`).
* Run GitHub actions using Java 8 and 11 to prove that both work (and to prevent regressions).
* Remove unnecessary Mockito stubbings (version of Mockito that supports Java 11 by default fails tests on these).
* Fixed Javadoc warnings in `circus-train-api` module.

### DIFF
--- a/circus-train-core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
+++ b/circus-train-core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
@@ -1,0 +1,1 @@
+mock-maker-inline
--- a/circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainParquetSchemaEvolutionIntegrationTest.java
+++ b/circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainParquetSchemaEvolutionIntegrationTest.java
@@ -16,7 +16,7 @@
 package com.hotels.bdp.circustrain.integration;
 
 import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
+import static org.hamcrest.MatcherAssert.assertThat;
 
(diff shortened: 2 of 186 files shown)


circus-train-core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
massdosage, line 1: This is the magic to allow Mockito to mock final classes (needed in `ManifestAttributesTest`) so we don't need PowerMock any more, for the curious - https://www.baeldung.com/mockito-final

circus-train-integration-tests/src/test/java/com/hotels/bdp/circustrain/integration/CircusTrainParquetSchemaEvolutionIntegrationTest.java
```diff
@@ -125,6 +125,7 @@ public void addFieldAtEnd() throws Exception {
         "2\tafter\t2"
     );
     runTest(schema, evolvedSchema, new FieldDataWrapper(), afterEvolution);
+    runDataChecks(evolvedSchema, expectedData);
```
massdosage, line 128: These assertions were missing so I added them (same on L142 below).

nvitucci (approved):
Discussed some details with @massdosage offline. Good to go for me.

andreeapad (approved):

PR #213, opened 2021-02-15
Added a fix for generating the target partition path if base and part…

### LINKED ISSUE #212: Replication (metadata update) fails when base path it not matched in all paritions
Sometimes replication fails with an error like:
```
Caused by: com.hotels.bdp.circustrain.api.CircusTrainException: Partition path 's3://bucket/db/table/1539359466542/acquisition_instant=20181012T155106Z' does not seem to belong to data source path 's3://bucket/table'
```
Or something similar. This is cause by a check in the code the tries to create the target partition location based on the location layout. If possible this should be fixed in the metadata of the source table. But since the data is actually copied it would be nice if Circus Train can make an estimate on what the path should look like.

patduin:
fixes #212

### DIFF
--- a/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
+++ b/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
@@ -35,7 +35,9 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.conf.HiveConf.ConfVars;
+import org.apache.hadoop.hive.metastore.Warehouse;
 import org.apache.hadoop.hive.metastore.api.ColumnStatistics;
+import org.apache.hadoop.hive.metastore.api.MetaException;
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.SetPartitionsStatsRequest;
 import org.apache.hadoop.hive.metastore.api.Table;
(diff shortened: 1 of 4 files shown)


circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
```diff
@@ -268,6 +271,29 @@ public void updateMetadata(
     }
   }
 
+  private Path createReplicaPartitionLocation(
+      TableAndStatistics sourceTableAndStatistics,
+      Partition sourcePartition,
+      ReplicaLocationManager locationManager) {
+    try {
+      return locationManager.getPartitionLocation(sourcePartition);
+    } catch (CircusTrainException e) {
```
patduin, line 280: This is thrown when base location doesn't match partition location, strangely enough the actual data copy doesn't have a problem with this so this logic fixes it by generating the partition name (instead of taking the name from the source partition folder).

barnharts4 (commented):
Looks good!

circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/replica/Replica.java
patduin: Hmm I'm actually a bit concerned this might not work correctly with how this is implemented: https://github.com/HotelsDotCom/circus-train/blob/main/circus-train-core/src/main/java/com/hotels/bdp/circustrain/core/source/HdfsSnapshotLocationManager.java#L124

I need to find out how the different copiers generate the target paths.

patduin:
Gonna close this this is not going to work correctly for some use cases (e.g. partition x is in bucket x and partition y is in bucket /y/z) and might lead to us not getting a notification and data being not in a different location than advertised in the Metastore. I'll update the issue with findings and rewrite it to better capture the requirements unfortunately that's a lot more work there is no quick fix imo

